PR #1, opened 2015-04-21
Gitignore

kejbaly2:
nothing major, just extended .gitignore file


psss:
Thanks for the patch, Chris. Do we really need all that stuff? It
was originally there as part of the template but I removed it. To
me it looks like 90% is just irrelevant.

kejbaly2:
.gitignore for sure is needed; i'll reduce it down to the essentials though and re-submit.

PR #3, opened 2015-04-21
Removes python-nitrate dependency

kejbaly2:
Includes everything from the X_nitrate branch too... so you can just merge this if you agree with X_nitrate and close the other pull request....

Logging (and utils) have been moved from python-nitrate to status-report/utils.py

User() now extracts name, email, login from config (and opts) only 'email' in the form of 
    "My Name" <joe@email.com>


kejbaly2:
looks good; currently in master. Thanks psss!

psss:
Thank to *you*, Chris! BTW, nice idea with handling the users with
the whole email string. I'm still a little bit thinking about --user
option for backward compatibility... but I'm starting to like your
solution much more. So we could probably sacrifice this... as "less
is more"... at least sometimes ;-)

PR #19, opened 2015-09-06
Refactor package dir path

kejbaly2:
`python setup.py develop` mode isn't working with the current source/module structure due to a known issue without resolution. 

Any reason against moving the sources out of the subdir and up to repo root? I've checked - docs are building, tests passing, rpm builds and setup.py with the new dir structure. Docker doesn't depend on the dir structure AFACT.


psss:
Thanks for the patch and detailed description of the problem.
In this way we get a bit more flat structure. I like this. Merged.

PR #20, opened 2015-09-06
_static holds images, etc; _templates hold templates

kejbaly2:
these shouldn't be ignored on git commits; they are static folders that are expected to contain files linked to / used / displayed from the sphinx rst files. _static hold images, css, etc; _templates hold template files. They're paths that get set-up in docs/conf.py; they should exist. Sphinx complains when they don't.

Strangely though, when i try to add the directories git still seems to be ignoring them even with the updated .gitignore... so I can't seem to actually get them added to HEAD. 

The update to .gitignore is a good idea anyway; we can work on adding the dirs later.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -30,8 +30,6 @@ pip-log.txt
 
 # sphinx docs
 _build
-_static
-_templates
 
 # rpm build files
 tmp/


psss:
Makes sense. Merged. Thanks.

PR #21, opened 2015-09-06
Feat.sphinx template

kejbaly2:
This doesn't have to fly... but I like the sphinx bootstrap theme. I've added a check that if it happens to be available then it'll be used and added 'extra' installers for 'docs' to setup.py

    python setup.py -e .[docs]

to install docs related dependencies, including this theme.


kejbaly2:
let me know if you don't care much for the boot strap theme. Minimally we should use 'sphinx_rtd_theme' instead of default; it's more modern, looks nicer and comes with latest version of sphinx. Also, it's the default that comes with read the docs, so local and rtd docs would be consistent.

kejbaly2:
wow i need to figure out how to use git; there are way too many commits here... going to try starting over.

psss:
Does this demo match the style you like?
https://ryan-roemer.github.io/sphinx-bootstrap-theme/#

kejbaly2:
yes

kejbaly2:
but i'm actually quite alright with the rtd format; what do you think?

kejbaly2:
demo: https://mrbob.readthedocs.org/en/latest/

psss:
> wow i need to figure out how to use git; there are way too many
> commits here... going to try starting over.

Perhaps a `git rebase master` on the feature branch could help?
Not sure whether the pull request can be updated in this way.

psss:
I like the default Read The Docs theme.
The demo you've provided is in the default or not?

kejbaly2:
move conversation over to the new pull request.

https://github.com/psss/did/pull/23

PR #22, opened 2015-09-07
mrbob templates for quickly creating new plugins

kejbaly2:
https://mrbob.readthedocs.org/en/latest/

first pass at a mrbob template for quickly creating a new plugin. 

The resulting plugin even comes with some default tests.

We'll want to add additional docs strings to the plugin to give helpful hints to the developers and maybe include addition methods that can be overridden with examples, etc. 

But for now, this is helpful as-is


kejbaly2:
I'm going to close this and clean up the commit history and submit a new pull request

PR #23, opened 2015-09-07
use bootstrap if installed; fallback:rtd, default

kejbaly2:
pulled and applied against master.

IF bootstrap is installed, use it; otherwise use rtd_theme (comes with sphinx in 1.3) and fallback to 'default' if neither is available

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -15,6 +15,21 @@
 import os
 import sys
 
+try:
+    # use bootstrap theme if user has it installed
+    import sphinx_bootstrap_theme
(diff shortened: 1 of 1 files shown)


kejbaly2:
the RTD theme comes installed with Sphinx 

http://sphinx-doc.org/theming.html

kejbaly2:
this way we can play with the bootstrap theme and see if we like it but only use it if we have it installed which would require us to install it explicitly to enable... so default will always be to use rtd theme unless we actively want to use bootstrap

psss:
OK, makes sense in this way.

PR #24, opened 2015-09-07
enable package 'extras' (dependencies) installation

kejbaly2:
In addition to the core requirements it's possible to define
extra package dependencies that can be installed along with
the core package + deps.

`install` usage::
    pip install did[tests,docs]

`develop` usage::
    python setup.py -e .[tests,docs]


kejbaly2:
merging to 0.4.1 branch

PR #25, opened 2015-09-07
mr.bob template to generate new default plugin

kejbaly2:
src: https://mrbob.readthedocs.org/en/latest/

Usage::
    mrbob $GITREPO/bobtemplates/plugin -O $GITREPO [-c ~/.mrbob.ini]

Where bobtemplate/plugin is the basedir for the plugin templates; all
files here will be copied there after variable replacement. `mrbob`
will ask questions to fill in variable values. You can save values
use into a mrbob.ini file and use that instead for non-interactive use.

The resulting plugin even comes with some default tests.

We'll want to add additional docs strings to the plugin to give helpful
hints to the developers and maybe include addition methods that can be
overridden with examples, etc.

But for now, this is helpful as-is

### DIFF
--- a/bobtemplates/hooks.py
+++ b/bobtemplates/hooks.py
@@ -1,0 +1,36 @@
+#!/usr/bin/env python
+# coding: utf-8
+# Author: "Chris Ward" <cward@redhat.com>
+
+'''
+Mr.Bob Hooks
+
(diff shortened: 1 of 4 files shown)


kejbaly2:
merging to 0.4.1 branch

PR #26, opened 2015-09-07
Pep8.fixes

kejbaly2:
clean-up of some remaining pep8 issues flake8 picks up and my vim environment with python-mode installed complains about constantly everytime i open these files.


kejbaly2:
merging to 0.4.1 branch

psss:
I don't think it's a good idea to adjust code just to make editor
happy when it is IMO nicely readable. Section comments separate
classes and provide 2 blanks lines suggested by PEP8 plus a little
bit more ;-) If there is no section comment it makes sense to have
2 blank lines.

I've pushed fix for the jira error handling issue as fd3999c.

kejbaly2:
tldr; pep8 suggests two \n even between comment and class definition.

It's not my editor per se, rather flake 8; which runs pep 8 checker (with
default settings).

https://pypi.python.org/pypi/flake8
https://pypi.python.org/pypi/pep8
https://github.com/klen/python-mode

http://pep8.readthedocs.org/en/latest/intro.html

Do you run flake8/pep8 checker in your dev env? (I recommend python-mode if
you're not already using it)

RE: 'section comments'  Since it seems your hesitant about the extra \n
after the header, what about an extra \n before the big headers? That would
work for me too (2 before, 1 after)


```
\n
\n
#
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
# Git Stats
#
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
\n

class ...
```

-Chris

psss:
OK, I've added extra lines before class comments. This does not
look too bad. However, the `NOQA` tags are just ugly :-)

PR #27, opened 2015-09-07
updated path to bin/did

kejbaly2:
fix spec and make file so rpm's are building correctly again.


kejbaly2:
merging to 0.4.1 branch

PR #31, opened 2015-09-07
New Features: Use invoke to run commands and build mrbob project templates

kejbaly2:
Added some basic project tasks: build, test, coverage, clean_git, etc. Usage eg

    cd $GITREPO
    invoke --list
    invoke --help build
    invoke build rpm

Also included are mrbob templates for: py_file, plugin

    invoke bob_did_plugin
    invoke bob_py_file

did_plugin builds a working module after you supply it with a few answers to some basic questions; includes basic tests too.
py_file produces a 'standard' python file for the project (header, py3 compatability future imports, etc)

### DIFF
--- a/bobtemplates/plugin/.mrbob.ini
+++ b/bobtemplates/plugin/.mrbob.ini
@@ -1,0 +1,36 @@
+[questions]
+
+plugin.name.question = Name of the plugin (one word; lower-case)
+plugin.name.required = True
+plugin.name.help = Should be something like 'rh_bugzilla'.
+
+plugin.app.question = Name of the App (CamelCase)
+plugin.app.required = True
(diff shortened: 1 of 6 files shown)


psss:
Hi Chris, I'm not sure about the invoke thing. It seems to be very
similar to what we can already achieve by Makefile targets. What's
the main point of adding this? I imagine `make new_plugin` could
serve similar purpose just fine.

psss:
> Yes, invoke is a python based makefile; please consider it; i
> don't think it hurts to have them both. I prefer working in python.

Here I would rather prefer less is more. I like to prevent
duplication wherever possible. The less code the less maintenance.
From what I can see we can handle all basic actions with Makefile.
I've added a couple more targets: `make test`, `make coverage`,
`make hooks` and updated the documentation in 1e4c85e.

Issue #45, opened 2015-09-24
Add statistics of CLOSED bugs for bugzilla plugin


xersin:
Statistics of closed bugs should be available too. Time spent by verification and closing of old/non-existing/non-resolvable bugs can be significant but valuable.

PR #46, opened 2015-09-24
Add statistics of closed bugs for bugzilla plugin

xersin:
Add statistics of CLOSED bugs for bugzilla plugin.
This resolves issue #45 .

### DIFF
--- a/did/plugins/bugzilla.py
+++ b/did/plugins/bugzilla.py
@@ -1,7 +1,7 @@
 # coding: utf-8
 
 """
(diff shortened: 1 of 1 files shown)


psss:
Thanks for the patch, Stanislav. I did a couple of adjustments: Limit the main search to bugs in closed status and ensure the status change has been performed by given user, otherwise it would should also bugs closed by automation when bugs are released. Also test suite failure has been fixed and a new test cases added.

PR #50, opened 2015-09-28
base.Config to local scope (private)

kejbaly2:
any reason not to use the locally scoped Config rather than re-calling it and creating new one-off instances over and over?


kejbaly2:
woops; tests are failing; nevermind. closing. will reopen later.

PR #52, opened 2015-09-30
Add the trello plugin

maybelinot:


psss:
Thanks for writing the Trello plugin, Eduard. Looks nice! Is there any public board which we could test it on? Could you include at least a minimal smoke test? Thanks.

Regarding the plugin order: I think it would be better not to modify order of existing plugins otherwise user adjustments could be affected. What about picking let's say 450 for trello?

Please consider using commit messages up to 50 chars so that they can be directly pasted into the changelog: http://did.readthedocs.org/en/latest/contribute/#commits

maybelinot:
@psss Thanks for the feedback. I tried to correct all the flaws and added the initial set of tests [here](https://github.com/psss/did/pull/55)

PR #53, opened 2015-10-01
upgrade to argparse

kejbaly2:
http://stackoverflow.com/questions/3217673/why-use-argparse-rather-than-optparse

    As of 2.7, optparse is deprecated, and will hopefully go away in the future.

Deprecated. I think it's clear optparse should be replaced with argparse. The migration is pretty simple. I think this covers everything. 

Sub commands are also a bit nicer in my experience.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -33,3 +33,6 @@ _build
 
 # rpm build files
 tmp/
+
+# vim autocomplete
+.cache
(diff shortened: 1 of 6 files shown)


psss:
Makes sense, thanks for the patch, Chris. We still need to adjust requires in the spec file as `python-argparse` is a separate package on RHEL6. Seems we will need to add some conditionals to limit the extra requires for python-2.6 only.

psss:
Conditional based on the dist tag should be sufficient:
https://fedoraproject.org/wiki/Packaging:DistTag
I will look into this later today.

psss:
I've committed a modified version of the patch as 9ec7df4.
Error handling has been improved in 21045e3.

PR #54, opened 2015-10-01
don't create new Logging instance uncessarily

kejbaly2:
Minor thing; why call utils.Logging("did") instead of just using the logger already cached into the namespace?

cli.py lines
    20 from did.utils import log
...
```
 39         # Enable debugging output (even before options are parsed)
 40         if "--debug" in sys.argv:
 41             utils.Logging("did").set(utils.LOG_DEBUG)
```

did.utils.log is utils.Logging("did").logger

What about

```
 39         # Enable debugging output (even before options are parsed)
 40         if "--debug" in sys.argv:
 41             log.setLevel(utils.LOG_DEBUG)
```

### DIFF
--- a/did/cli.py
+++ b/did/cli.py
@@ -38,7 +38,7 @@ class Options(object):
 
         # Enable debugging output (even before options are parsed)
         if "--debug" in sys.argv:
-            utils.Logging("did").set(utils.LOG_DEBUG)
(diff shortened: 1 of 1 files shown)


psss:
Makes sense, thanks for the patch.

PR #55, opened 2015-10-02
Add the trello plugin

maybelinot:


kejbaly2:
Let's take a look at those errors next week; the 'user' issues are related to a change made to the Stats class API; you need to update your Stats __init__ to accept this new argument.

psss:
Great! Thanks for incorporating the feedback, Eduard. After a couple of adjustments in 285f07e I've merged the plugin into master and the tests look happy ;-) Here's a couple of thoughts that appeared to me when I was going through the code:

As for now the url is hardcoded. I think it would be better to have this configurable (even if there is just one instance now).

What about calling `board_links` argument just `boards`?

Would it be possible to initiate the session connection only upon the first request? Now even `did -h` does the session setup and this is unnecesary (and a little bit slow as well).

I've done a couple of changes to the option handling as they were not displayed correctly:

    Trello stats group:
      --toolscreateCard   Actions in Cards
      --toolsupdateCard   Actions in Cards
      --toolsupdateCard:closed
                          Actions in Cards
      --toolsupdateCard:idList
                          Actions in Cards
      --toolsupdateCheckItemStateOnCard
                          Actions in Checklists
      --tools             All above

Now it's a bit better: Header mentions the section name and options describe what they are for:

    Trello updates for tools:
      --tools-createCard  Trello cards created
      --tools-updateCard  Trello cards
      --tools-updateCard:closed
                          Trello cards closed
      --tools-updateCard:idList
                          Trello cards moved
      --tools-updateCheckItemStateOnCard
                          Trello cards
      --tools             All above

But still, they are a bit confusing. What about having something like this?

    --tools-cards-created
    --tools-cards-closed
    --tools-cards-moved

Last thing, all config sections support email/login aliasing and user using `email` and `login` as the key. Would it make sense to user `login` instead of `user` in trello plugin as well?

Once again, thanks for writing the plugin and the tests as well!

PR #57, opened 2015-10-03
New plugin: bit.ly links saved

kejbaly2:
bitly deps; date->epoch convertion; tests;
docs; error handling; use int()s;
includes info about getting oath tokens


kejbaly2:
closing; need to fix something; brb.

psss:
What about just pushing an updated version of the patches to the same branch / pull request instead of closing them and opening new ones? This should work nicely with `git commit --amend` as well I guess.

Issue #41, opened 2015-09-22
git repos expansion containing non-repo fails


dahaic:
When path I specify as keeper of git repos [i.e. with /* expansion] contains non-repo directory, it fails with error complaining about GIT_DISCOVERY_ACROSS_FILESYSTEM not set.
```
 ERROR  fatal: Not a git repository (or any parent up to mount point /home)
 Stopping at filesystem boundary (GIT_DISCOVERY_ACROSS_FILESYSTEM not set).
 ERROR  Unable to check commits in '/home/dahaic/RH/git/tests/UPSTREAM'
```
If env variable is set, it fails nevertheless.
```
 ERROR  fatal: Not a git repository (or any of the parent directories): .git
 ERROR  Unable to check commits in '/home/dahaic/RH/git/tests/'

```
It would be nice to skip such directory and not fail.

psss:
Agreed, it probably makes sense to give a warning in such case.

PR #61, opened 2015-10-04
fixing Unable to access git repo

glum23:
Similar problem as described in #41 when path contains a file (not repo-directory).
Original error:
```
 ERROR  Unable to access git repo '/home/user/tests/git/file'
```


psss:
I think it probably makes sense to completely ignore files in this case.

PR #63, opened 2015-10-05
Enable use of requests[-kerberos] if installed

kejbaly2:
Installation

```
pip install -e .[requests]
```

Enable use of requests[-kerberos] if installed
Fallback to using urllib2 if requests dependencies aren't available
Requests depedencies included as extras in setup.py
Enabled ssl_verify setting in config
New helper util added to convert untyped ConfigParser bool values

### DIFF
--- a/did/plugins/jira.py
+++ b/did/plugins/jira.py
@@ -33,15 +33,25 @@ Notes:
 
 from __future__ import absolute_import, unicode_literals
 
-import re
-import json
-import urllib
-import urllib2
-import cookielib
-import dateutil.parser
-import urllib2_kerberos
+try:
+    import requests as rq
+    from requests_kerberos import HTTPKerberosAuth, DISABLED
+    import warnings
+    # Output warnings ONE time, then supress
+    warnings.simplefilter('default')
+    has_requests = True
+except ImportError:
+    has_requests = False
+    import urllib2
+    import urllib2_kerberos
+    import json
 
-from did.utils import log, pretty, listed
+import cookielib
+import urllib
+import re
+import dateutil.parser
+
+from did.utils import log, pretty, listed, as_bool
 from did.base import Config, ReportError
 from did.stats import Stats, StatsGroup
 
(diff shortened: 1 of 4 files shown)


kejbaly2:
ok, again, urllib2 solution does not work for me; python-requests does. simple as that. To avoid upsetting people with the extra deps though, i've made it optional. If requests depedencies aren't installed it will fall back to the original urllib2 implementation.

psss:
Thanks for putting together the pull request Chris. After today's IRC discussion and (for me) the new fact that we already depend on `python-requests` through the dependency on `python-bugzilla` (which requires the module) I think we should probably just go forward with requests and save ourselves from complicating the code with the conditional stuff. Unless @mcepl has any further strong objections. But I think it just does not make sense to have the library at hand and use more complicated approach. Plus I don't think removing `python-bugzilla` from the dependencies is an option here.

kejbaly2:
OT: regarding python-bugzilla: i think it should be moved out of the hard
dependencies and into extras.

mcepl:
I really don’t want to get into disputes: but could some explain here briefly what's the problem with the authentication which you try to workaround in `_ssl_handler`? If it is need to be authorized before the first attempt to connect (which breaks the way Python correctly according to the spec with HTTPS), then I have dealt with this extensively in https://gitlab.com/mcepl/urllib2_prior_auth/blob/master/urllib2_prior_auth.py (the package on PyPI `urllib2_prior_auth`) and it could be in the end rather easily resolved with the plain `urllib2`. Also see http://bugs.python.org/issue19494 (my solution is in python 3.5).

kejbaly2:
Actually, I couldn't work around it with _ssl_handler either. 

I imagine there are other solutions; requests works for me (i use it in other projects) and now it's a dependency through python-bugzilla anyway, so.... how would we clear that situation up? If we pull requests in anyway through python-bugzilla, i'd say using requests is ok anywhere else too. 

I also don't see any reason not to leave the urllib compatability code in their too to avoid making requests a hard dependency on did itself, but only on the plugin. That works too.

I'm sure Petr will figure something out. :+1:

psss:
We are now using `requests_gssapi` for Jira stats. See b61d50b. Closing

PR #64, opened 2015-10-05
Updated activity descriptions; use default filters

kejbaly2:
Updated the activity descriptions (api doc) to be (IMO) more readable.
Add default set of filters to use if not defined in user config.


psss:
Thanks for the patch, merged after a couple of adjustments in bb7a81c.

PR #65, opened 2015-10-06
Add New Plugin: Idonethis

kejbaly2:
Query all your 'dones' on idonethis.com


kejbaly2:
Last run is failing in the github tests; likely we've hit some API usage limit.

psss:
Thanks for the plugin, Chris. I just renamed the `apitoken` option to `token` in d6a8a45 to make it consistent with the rest of the plugins and slightly adjusted error reporting. Also one of the tests was failing because of getting a different date from the server.

PR #66, opened 2015-10-06
New plugin: bit.ly links saved

kejbaly2:
* bitly deps; date->epoch convertion; tests;
* docs; error handling; use int()s;
* includes info about getting oath tokens
* install the extra bitly api dependency in travis-ci


psss:
Thanks for the new plugin, Chris. I did a couple of adjustments 0c3f91a and merged.

Issue #60, opened 2015-10-03
Date range might be a little bit confusing


psss:
Internally `options.since` and `options.until` dates are interpreted as the midnight of the respective date. This means that plugins need to query like this:

    date >= options.since and date < options.since

This might be a little bit confusing at the first sight. But might be also useful as some services internally handle queries as datetime so for these you would have to say:

    date > options.since and date < options.since

So the question is: Shall we leave it as it is? Or make the `until` option content match what is given on the command line? (Now `--until 2015-10-03` is saved as `2015-10-04`.)

PR #71, opened 2015-10-12
Use UTC datetimes for Date()

kejbaly2:
Changed the default behaivor of Date() to
normalize everything to UTC timezone datetime.datetime
instances instead of datetime.date to maintain the
datetime resolution for cases where it matters.

Also, while the default unicode(Date()) outout is
YYYY-MM-DD, the format can now be changed on demand
using the fmt variable. eg (from tests/test_base.py)

```
from did.base import Date

d = '2015-01-01'
dt = '2015-01-01 00:00:00'
dtz = '2015-01-01 00:00:00 +0000'
dtz_cet = '2015-01-01 00:00:00 +0200'
# since we're 2 hours ahead, in UTC we're -2
dtz_cet_utc = '2014-12-31 22:00:00 +0000'
d_cet_utc = '2014-12-31'

# Default format will return the date in YYYY-MM-DD form
# after date conversion from defined timezone to UTC, if
# timezone is set on the incoming date value.
assert str(Date(d)) == d
assert str(Date(dt)) == str(d)
assert str(Date(dtz)) == str(d)
assert str(Date(dtz_cet)) == str(d_cet_utc)

# different formats should all work as strftime() expects
fmt = '%Y'
assert str(Date(d, fmt=fmt)) == '2015'
fmt = '%Y-%m-%d'
assert str(Date(d, fmt=fmt)) == d
fmt = '%Y-%m-%d %H:%M:%S'
assert str(Date(dt, fmt=fmt)) == dt
fmt = '%Y-%m-%d %H:%M:%S %z'
assert str(Date(dt, fmt=fmt)) == dtz
assert str(Date(d, fmt=fmt)) == dtz

# unicode works as expected
assert unicode(Date('2015-01-01')) == unicode(d)

# specify timezone
iso = '%Y-%m-%d %H:%M:%S %z'
assert unicode(Date(dtz_cet, fmt=iso)) == dtz_cet_utc
```

Requires one new dependency - pytz http://pythonhosted.org/pytz/

Fixed some misc tests that depended on old Date features which
needed to be migrated.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -56,6 +56,12 @@ format ``Name Surname <email@example.org>`` to display full name
 in the report output. For date values ``today`` and ``yesterday``
 can be used instead of the full date format.
 
+Note, all dates are assumed to be UTC unless otherwise specified. 
+ISO format dates can be used to specify otherwise. eg, 
+
+ * midnight CET is `2014-12-31 22:00:00 +0000` UTC (10pm)
+ * midnight UTC is `2015-01-01 02:00:00 +0200` CET (2am)
+
 --email=EMAILS
     User email address(es)
 
--- a/did/base.py
+++ b/did/base.py
@@ -9,13 +9,13 @@ import re
 import sys
 import codecs
 import datetime
-import optparse
 import StringIO
-import xmlrpclib
 import ConfigParser
 from dateutil.relativedelta import MO as MONDAY
 from ConfigParser import NoOptionError, NoSectionError
 from dateutil.relativedelta import relativedelta as delta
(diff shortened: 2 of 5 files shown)


kejbaly2:
travis fails are because of github tests failing which are likely because of us hitting some API call limit...

psss:
Currently there does not seem to be a clear need for handling time zones and exact times. We might revisit that later. Perhaps in connection with #60. Also something like `did last hour` could be handy. But that would require many changes in individual plugins. Closing for now.

Issue #79, opened 2015-12-03
RFE: Support for did yesterday


rbarlow:
I can do `did --since <yesterday's date>`, but it would be sweet if this supported `did yesterday`.

PR #85, opened 2016-01-21
Added support for 'did yesterday' [#79]

mfrodl:


psss:
Thanks for the patch, Martin. I've merged a duplicate pull request #91 which basically does the same but in addition includes a test.

PR #94, opened 2016-04-05
Allow user to set a specific username for GitHub plugin

happz:
It may happen that name of your GitHub account is not the same as the
user you set in [general] section of your config file. In such case,
GitHub would not have any stats for you, would it? Tell GitHub plugin to
use the correct username.

### DIFF
--- a/did/plugins/github.py
+++ b/did/plugins/github.py
@@ -8,9 +8,13 @@ Config example::
     type = github
     url = https://api.github.com/
     token = <authentication-token>
+    username = <username>
 
 The authentication token is optional. However, unauthenticated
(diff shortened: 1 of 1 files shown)


happz:
Argh! Reading tests/plugins/test_github.py, and there it is already solved... :/ Never mind then...

psss:
Yes, providing different login/email for particular stats is supported for all plugins. This should be documented on the following two links:

http://did.readthedocs.org/en/latest/config/#aliases
http://did.readthedocs.org/en/latest/modules/#did.base.User

Let me know if the current documentation is not sufficient. Thanks.

PR #97, opened 2016-05-24
Add `login` key to github section in example config

AloisMahdal:
example/config does not illustrate (common) possibility to override
login for particular plugin.

Github is one of those that may likely need this, so let's add it there.


sochotnicky:
Yeah, I had to look at the code to figure this one out myself :-) :+1:

psss:
I see. Makes sense. I'll add it to the module example as well. Thanks for the patch, Alois.

PR #102, opened 2016-06-30
Fix typo in mr.bob example

AloisMahdal:
According to [mr.bob documentation](https://mrbob.readthedocs.io/en/latest/userguide.html) and my experiment, the pip package
name is `mr.bob`, not `mrbob`.


psss:
Yeah, that was a typo. Merged. Thanks for the fix, Alois.

Issue #33, opened 2015-09-21
'a' recognized as valid argument


lukaszachy:
Allowed arguments are described in usage string as 
Usage: did [last] [week|month|quarter|year] [opts]

I would expect failure for each unrecognised positional argument. But that is only partially true:
$ did b
ERROR  Invalid argument: 'b'

$ did a
Status report for this week (2015-09-21 to 2015-09-27).
...

AFAIK there should be no positional argument which 'a' matches.

PR #106, opened 2016-08-09
Show weeknum in status-report for week period

xjezda00:
Petře,

Following pull request adds ISO week number to this/last week period header for better orientation, for example:

$ did this week
Status report for this week # 32 (2016-08-08 to 2016-08-14).

$ did last week
Status report for the last week # 31 (2016-08-01 to 2016-08-07).

Regards,
DJ

### DIFF
--- a/did/base.py
+++ b/did/base.py
@@ -302,10 +302,10 @@ class Date(object):
         else:
             if "last" in argument:
                 since, until = Date.last_week()
-                period = "the last week"
+                period = "the last week # " + since.datetime.strftime("%V")
             else:
                 since, until = Date.this_week()
-                period = "this week"
+                period = "this week # " + since.datetime.strftime("%V")
         return since, until, period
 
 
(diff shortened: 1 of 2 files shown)


psss:
Thanks for the idea, David. This looks fine, however I would prefer if the feature is configurable. There could be an option in the general section for turning the week numbers on or off. Or, perhaps, even better we could allow simple custom header formatting in the general config section using something like this:

```
header = Status report for {relative_week} ({since} to {until}).
```

For your example it could looke like this:

```
header = Status report for the week #{week} ({since} to {until}).
```

This could bring some more flexibility also for generating monthly reports:

```
header = Status report for {month} {year}.
```

The resulting headers would then look like this:

> Status report for this week (2016-11-07 to 2016-11-13).
> Status report for the week #33 (2016-11-07 to 2016-11-13).
> Status report for November 2016.

We would have to find a way how to set different output for week/month/quarter/year. Perhaps a separate config section could solve this. What do you think?

psss:
Thinking about it a bit more, it seems to me now the proposal above is just too complicated. Let's do it much more simply by showing month names and week numbers by default. This is more useful than the relative reference. Fixed in 8682a35d1cf83f350081369e51e3455d335d9618.

PR #108, opened 2016-10-04
Deduplicate gerrit items when merging

sochotnicky:
When defining account with multiple aliases it happens that gerrit reports the
same item for both aliases. This change fixes this behaviour by using set()
instead of list and thus deduplicating items. To make this work we define **eq**
and **hash** functions in Change Gerrit class.

This could be useful for other plugins as well - when multiple people work on
the same item it is (presumably) enough to report it once and not once for each
person when merging.


psss:
Makes sense. Thanks for the patch, Stano.

PR #109, opened 2016-10-25
Report bugs created with Patch keyword as patched

mfrodl:
Currently, `PatchedBugs` only reports previously existing bugs to which the Patch keyword was added. Bugs to which the reporter added the keyword during their creation are not reported. By applying this commit, `PatchedBugs` will report such bugs as well.

### DIFF
--- a/did/plugins/bugzilla.py
+++ b/did/plugins/bugzilla.py
@@ -490,6 +490,38 @@ class PatchedBugs(Stats):
                 query, options=self.options)
             if bug.patched(self.user)]
 
+        # When user adds the Patch keyword when creating a bug, there is no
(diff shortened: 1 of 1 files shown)


mfrodl:
Also, note that I had to use the new XMLRPC argument format (`f1`, `o1`, `v1` etc.) because the old one which is currently used in did (`field0-0-0`, `type0-0-0`, `value0-0-0` etc.) does not have sufficient expressive power. Most notably, the old style does not provide a way to express negation (keywords *not* changed to "Patch").

In order to prevent mixing old and new syntax, we should consider rewriting all the existing queries to use the new style as well.

PR #111, opened 2016-10-26
Use new argument syntax in Bugzilla XMLRPC queries

mfrodl:
The old field names (`field0-0-0`, `type0-0-0`, `value0-0-0` etc.) have been replaced with the new ones (`f1`, `o1`, `v1` etc.). The old syntax was unable to express more complex queries which may be needed sometimes; for instance see #109.


psss:
Thanks for the patch, Martin!

PR #112, opened 2016-10-26
Fixed test for Trac plugin

### LINKED ISSUE #110: Failing test for Trac plugin
The Fedora Infrastructure project, previously using Trac for activity tracking, has been migrated to Pagure. As a consequence, the test for Trac plugin started to [fail](https://travis-ci.org/psss/did/builds/164860748).

We need to modify the test to use another, still active Trac project. It can be for instance one of these:

https://fedorahosted.org/fedora-apac/
https://fedorahosted.org/design-team/

mfrodl:
Fixes #110.

### DIFF
--- a/tests/plugins/test_trac.py
+++ b/tests/plugins/test_trac.py
@@ -14,8 +14,8 @@ import did.base
 CONFIG = did.base.Config.example() + """
 [trac]
 type = trac
-url = https://fedorahosted.org/fedora-infrastructure/rpc
-prefix = FI
+url = https://fedorahosted.org/design-team/rpc
+prefix = DT
 """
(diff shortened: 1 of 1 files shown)


psss:
Thanks for the fix of the test, Martin. It seems we'll probably have to find another Trac instance at the beginning of 2017 as fedorahosted is planned to be retired:

> We are planning to retire fedorahosted.org on 2017-02-28. See the announcement at:
> https://lists.fedoraproject.org/archives/list/announce@lists.fedoraproject.org/thread/RLL3LFUPLYMAUKGZ5B3O64XKJXBT24KZ/
>

Issue #90, opened 2016-02-29
GitHub plugin Unicode char


hasys:
If commit message contain some special Unicode symbol (for example ' symbol) exception is thrown and it is not possible to create report. 

did run with --debug option:

> Traceback (most recent call last):
>   File "/usr/bin/did", line 36, in <module>
>     did.cli.main()
>   File "/usr/lib/python2.7/site-packages/did/cli.py", line 191, in main
>     user_stats.check()
>   File "/usr/lib/python2.7/site-packages/did/stats.py", line 135, in check
>     stat.check()
>   File "/usr/lib/python2.7/site-packages/did/stats.py", line 135, in check
>     stat.check()
>   File "/usr/lib/python2.7/site-packages/did/stats.py", line 89, in check
>     self.show()
>   File "/usr/lib/python2.7/site-packages/did/stats.py", line 103, in show
>     utils.item(stat, level=1, options=self.options)
>   File "/usr/lib/python2.7/site-packages/did/utils.py", line 76, in item
>     eprint(u"{0}\* {1}".format(u" " \* indent, shorted(unicode(text), width)))
>   File "/usr/lib/python2.7/site-packages/did/plugins/github.py", line 86, in **unicode**
>     unicode(self.id).zfill(PADDING), self.data["title"])
> UnicodeEncodeError: 'ascii' codec can't encode character u'\u2019' in position 36: ordinal not in range(128)

Problem git commit message is:

> GUVNOR-2415: Boundary events lose it’s documentation after save and reopen Business Process

hasys:
Did version:

> Name        : did
> Arch        : noarch
> Epoch       : 0
> Version     : 0.8
> Release     : 1.fc23
> Size        : 267 k
> Repo        : @System
> From repo   : psss-did
> Summary     : What did you do last week, month, year?
> URL         : https://github.com/psss/did
> License     : GPLv2+
> Description : Comfortably gather status report data (e.g. list of committed
>             : changes) for given week, month, quarter, year or selected date
>             : range. By default all available stats for this week are reported.

PR #113, opened 2016-10-27
Fixed Unicode bug in GitHub plugin

mfrodl:
Resolves issue #90; test included.


psss:
Thanks for the patch, Martin!

Issue #93, opened 2016-03-15
Github plugin: differentiate issues and pull requests


dkutalek:
According to:
https://developer.github.com/v3/issues/

Note: In the past, pull requests and issues were more closely aligned than they are now. As far as the API is concerned, every pull request is an issue, but not every issue is a pull request.

This endpoint may also return pull requests in the response. If an issue is a pull request, the object will include a pull_request key.

psss:
Martin's pull request has been merged to master. Closing.

PR #114, opened 2016-10-27
Separate pull requests from issues

mfrodl:
Resolves issue #93.


psss:
Thanks for the patch, Martin. Tests are now passing after a couple of adjustments needed because of the tight GitHub API limits for unauthenticated searches.

Issue #82, opened 2016-01-08
Any plan to connect to Google account and extract meetings from google calendar?


eduardocerqueira:
Hi there,

any plan to extend did to connect to a google account and extract meetings from Google Calendar?

mfrodl:
I have started working on a Google Apps plugin and will present my draft proposal soon.

Issue #11, opened 2015-06-22
email status report


kejbaly2:
Would be useful for me if I could get status report details for my email usage. 

    Emails Sent (new, reply) - date: subject (like git commits)

Starting with support for gmail using their api would probably be reasonable.

I'll work on this.

sferich888:
Are there any plans to work on or address this issue?

psss:
Hi, no plans as for now. If anyone would be interested to look into this, pull requests are definitely welcome.

PR #115, opened 2016-11-06
Report Google Calendar events

mfrodl:
Resolves issue #82.

This allows user to report events he or she attended or organized within the specfied time range based on their Google Calendar. It has been designed a way that allows easy extension to other Google Apps statistics in the future (for instance issue #11).

A few things may need further discussion:

* Should all events from user's calendar be reported or only the non-private ones? To go further, should there be a configuration option to choose between these two alternatives?

* What other calendar statistics could users want to report?

* What should be the default order with respect to other plugins? I set it to `50` because I would prefer having this section near the top of the report. Might be other users will differ.


mfrodl:
To be able to add test coverage for this plugin, we need a Google account. I propose creating a brand new one just for did testing purposes.

psss:
Thanks for working on this, Martin. Looks very good to me at the first sight! A couple of thoughts:

 * Could you please add a short instructions describing how to create the id & secret?
 * Regarding the default order, 50 sounds reasonable to me. Users can adjust this in config if they desire.
 * Agree with creating a separate Google account for testing. I can take care of this.

Will we be able to fully automate the testing in Travis while somehow keeping the credentials safe?

mfrodl:
The instructions for creating a pair of credentials are as follows (I will add them to the plugin's documentation):

1. Open https://console.developers.google.com/flows/enableapi?apiid=calendar
2. Select 'Create project' in the drop-down menu and click 'Continue'
3. Click 'Go to credentials'
4. In 'Where will you be calling the API from?' drop-down menu, choose 'Other UI (e.g. Windows, CLI tool)'
5. In 'What data will you be accessing?', choose 'User data'
6. Click 'What credentials do I need?'
7. Input 'did credentials' in the 'Name' field and click 'Create client ID'
8. In 'Product name shown to users', type 'did'
9. Click 'Continue', then 'Done'
10. Click the 'did credentials' link to display the credentials

Miscellaneous notes:

* PyPI package `google-api-python-client` needs to be added to dependencies
* The plugin always requires user to grant (read-only) access to his or her calendar. This is okay for now but if e.g. Gmail support is added to the plugin in the future and user will want to report emails only, access rights to calendar should not be required.

psss:
Great! Thanks for the detailed instructions, Martin.

dkutalek:
Nice one!

For testing this plugin, I have visited very relevant PyCon talk by Miro Hroncok:
https://cz.pycon.org/2016/speakers/talks/#miro-hroncok

You can consider using Betamax to record http requests/responses
of usage to be tested and then it can be replayed/retested 
without secrets to account in test itself (until Google API changes).

https://pypi.python.org/pypi/betamax/0.1.6

mfrodl:
Thanks for a useful tip, David. It might make sense to test this plugin in a mock environment rather than an actual Google Calendar instance.

It is also possible to provide the plugin with credentials (`client_id` and `client_secret`) which only allow a read-only access to the calendar. Any additional access rights on top of that (e.g. read-write access to calendar, access to Gmail etc.) would need to be explicitly allowed by the owner of the testing Google account. In this way, the Google account would still be safe from attackers who don't know the password – which will not be present in the test.

In theory, there is still the risk of attackers using the public credentials to exhaust the API quota on read-only access to the account. Should this concern prove justified, I will redesign the test to use a mock environment such as Betamax. Until then, I would prefer testing against a live account.

I will let you know once the plugin and test are ready for review.

mfrodl:
The plugin has been rewritten to simplify integration with other Google apps if needed in the future. Both to-do items from https://github.com/psss/did/pull/115#issuecomment-259155931 should be addressed by now. I also created a Google account (letstestdid at gmail.com) for testing purposes and wrote two test cases for the plugin. 

The pull request is ready to be reviewed and merged. Petr, please review the code and share your thoughts.

psss:
Looks nice and works great! I just did some little adjustments to make the online docs building working. Thanks for the plugin, Martin.

macrozone:
any chance to get gmail support as well?

PR #123, opened 2017-01-27
split Issues and PullRequests in the GitHub plugin

evgeni:
while issues and PRs are semantically almost identical, it is quite
interesting to split them in the report


evgeni:
ups, that is almost identical to #114 - I should have looked first

psss:
Thanks for the patch. Yes, #114 resolves this as well. Plus it includes test coverage. I'll close this one.

PR #124, opened 2017-01-27
add did.spec to MANIFEST.in

evgeni:
we need did.spec in setup.py, but we did not ship it in sdist

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,0 +1,1 @@
+include did.spec


psss:
The spec file is only needed for building the wheel. Why do we need to ship it?

evgeni:
`setup.py` cannot be executed without `did.spec`, which means e.g. the generated tarball (via `python setup.py sdist`) cannot be installed later with `python setup.py install`. This is e.g. needed when trying to run tests with `tox` in an venv.

psss:
I see. Thanks for explanation. Merging...

psss:
Merged in 4399e943ce7b5672d4a9689d9903693b4fcc1936.

Issue #129, opened 2017-03-02
gitlab support


mhradilek:
When working on projects managed with gitlab, It would be useful to have a plugin for tracking activities on gitlab.

Off the top of my head:
 * issues (created, commented, resolved)
 * commits
 * merged pull requests
 * wiki edits

thrix:
Yeah, we definitelly need this feature ...

evgeni:
please test https://github.com/psss/did/pull/130

PR #130, opened 2017-04-21
add basic GitLab support

evgeni:
fixes: #129

### DIFF
--- a/did/plugins/gitlab.py
+++ b/did/plugins/gitlab.py
@@ -1,0 +1,312 @@
+# coding: utf-8
+"""
+GitLab stats such as created and closed issues
+
+Config example::
+
+    [gitlab]
+    type = gitlab
+    url = https://gitlab.com/
+    token = <authentication-token>
+    login = <username>
+
+The authentication token is required.
+Use ``login`` to override the user associated with the token.
+See the :doc:`config` documentation for details on using aliases.
+
+__ https://docs.gitlab.com/ce/api/
+
+"""
+
+import requests
+import dateutil
+import itertools
+
+from did.utils import log, pretty, listed
+from did.base import Config, ReportError
+from did.stats import Stats, StatsGroup
+
+GITLAB_SSL_VERIFY = True
+GITLAB_API = 4
+
+# Identifier padding
+PADDING = 3
+
+# ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+#  Investigator
+# ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+
+class GitLab(object):
+    """ GitLab Investigator """
+
+    def __init__(self, url, token, ssl_verify=GITLAB_SSL_VERIFY):
+        """ Initialize url and headers """
+        self.url = url.rstrip("/")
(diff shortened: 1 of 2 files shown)


psss:
Thanks for the patch, Evgeni. Looks fine. Just a few comments: Could you include a couple of tests for the new plugin? (Is there a public gitlab instance which could be used?) I would recommend to always avoid one-letter variable names and rather use format() instead of "%" operator for text formatting.

Last two comments are just a nitpicks from general recommendations to keep stuff consistent:

 * When using hanging indent, the first line should be empty.
 * Commit: Complete sentence with the first word capitalized.

See also: http://did.readthedocs.io/en/latest/contribute/

evgeni:
There is gitlab.com which we could use I guess. But all API requests require authentication.
If you would create a user and share the token I could use that user for the tests.

thrix:
@evgeni thanks for working on this, could you make GITLAB_API configurable? I need version 3 ...

evgeni:
@thrix sadly v3 API has no user events endpoint, so while I could, it would be worthless :(

evgeni:
@psss added tests, but they obviously fail w/o a token :(

psss:
That's great! Thanks, @evgeni. Sorry, I don't have much time today but I'll create the account next week. Then we can make the tests working.

thrix:
@evgeni thanks, that is quite unfortunate :( we have just v3 available, anyway, thanks for the info

evgeni:
@thrix sadly the search API of GitLab is not as feature-rich as GitHub, so you cannot search for stuff like "all issues created by user X in the timeframe Y-Z", thus I used the events api, which provides a nice stream of data for that.

And I feel with you, one of the GitLab instances I use is also v3 only, so it does not work there :(

Ref: https://docs.gitlab.com/ce/api/issues.html#list-issues

evgeni:
@psss ping ;)

psss:
Oh! Sorry, @evgeni! My last weeks were quite busy...
Here's the token for testing: x8mA6GVs3Dw2zXMYMoc8
Do you need anything else to be set up on the GitLab server?

evgeni:
@psss cool, thanks! could you share the original account details (user/pass) too? then I can set up a test repo with test issues etc using the webUI instead of the API, which is easier.  (encrypt the data with gpg if you want to: C575A957E819BA18BF07E766A1B09B42333961E8 or E2FF 187B 5236 03AD 65D9  814B 81D0 CC18 450C E7AD)

psss:
Sure, here you are:

Username: did.tester
Password: :mH7K8RyjE2W}k6m

We can change the password when the repo is ready.

evgeni:
test failures are outside of test_gitlab.py :P

psss:
Thanks for the patch, Evgeni!

todoleza:
Thanks for the plugin. However I am unable to run the gitlab plugin against older {"version":"8.14.4","revision":"3ea6c8d"}, gitlab.com currently runs {"version":"9.4.0-rc3-ee","revision":"8ebdc5e"} with /api/v4 support.
I currently lack time to add backwards compatibility for /api/v3. Is there anyone willing to extend the plugin? Should I create different ticket for it?

evgeni:
@todoleza I think a separate ticket is in order. I am personally not willing/interested to work on v3 support, but maybe others are.

thrix:
do you mean reimplement activity log in the did module?

todoleza:
Hello @thrix, I wasn't how big the reimplementation, if any, would be needed. Anyway, the instance I am using is said to be going upgraded soon, so there's small, if any, need to proceed here.
I am not creating separate ticket then. Thanks

Issue #77, opened 2015-12-02
New plugin: Redmine


bowlofeggs:
I'd like to request support for a new plugin for the Redmine issue tracker:

http://www.redmine.org/

Thanks, did is very cool!

rbarlow:
This library may be useful for this: http://python-redmine.readthedocs.org/

PR #135, opened 2017-06-13
Add basic Redmine support

evgeni:
Fixes: #77 

Not using http://python-redmine.readthedocs.org/ because [the API does not expose activities](http://www.redmine.org/issues/22109)


evgeni:
test failures outside of the redmine plugin (idonethis, trac)

evgeni:
@psss ping? :)

psss:
Thanks for the plugin. And sorry for the delay.

PR #137, opened 2017-06-15
Pass after and before dates to GitLab API

evgeni:
This is supported since 9.3 and should allow us fetching smaller result sets.


psss:
Nice improvement. Thanks for the patch.

PR #139, opened 2017-08-03
Gitlab fixes

evgeni:
@psss these are two quite important fixes for the GitLab integration


evgeni:
@psss ping? :)

psss:
Thanks for the patch.

PR #141, opened 2017-09-12
Plugin for Sentry

Madeyro:
I created plugin for Sentry.io

They do not have API for requesting on activity, but it can be done with some workaround.
Please point out mistakes or improvements.


Madeyro:
All my tests  have passed.  
It's the other plugins before me, which have failed tests.

psss:
Thanks for the plugin, Maros! It looks very good. I've cleaned errors from other plugins but the test suite still fails (on sentry). Could you please have a look at the error?

 * https://travis-ci.org/psss/did/jobs/287673644

It looks like the intended test query did not go well (index out of range). You can base future commits on the devel branch which already includes sentry. Thanks.

Madeyro:
Thank you for your review.
The issue was caused by removal of one of the test cases and not updating the rest of them.
It should be OK now.

psss:
Thanks for the fix. All tests are now passing. Merged to master in 81294d5.

PR #142, opened 2017-10-20
Skip not unique issues

Madeyro:
Do not add to stats if the issue is already appended.
For example multiple comments on one issue.


Madeyro:
The decrease of test coverage is expected because the plugin is skipping issues which was already processed.

Issue #146, opened 2018-02-19
Test suite takes forever to run


frozencemetery:
I bailed about eight minutes in.  There are some timeouts - I'm not sure why it takes so long beyond that, but it does mean that it's infeasible for contributors to run it before code submission.

psss:
Test have been udpated. I think this should now be better.
The whole test suite now runs about 4-5 minutes:
https://travis-ci.org/psss/did/builds

Issue #145, opened 2018-02-19
fedorahosted trac URLs attempted access during tests


frozencemetery:
Several tests are failing due to fedorahosted trac acccess in the test suite.  It seems to really want to hit `https://fedorahosted.org/design-team`.

psss:
Unfortunately there is currently no instance to test against.
Trac tests have been temporarily disabled 77d577e.

PR #147, opened 2018-02-19
Port to python-gssapi

frozencemetery:
As part of this, migrate any urllib2_kerberos code to urllib-gssapi.

Similar caveats to [psss/python-nitrate#8](https://github.com/psss/python-nitrate/pull/8) apply:

- The test suite for this tool cannot reasonably be run; see #146  and #145 
- There are libraries ([requests-gssapi](https://github.com/pythongssapi/requests-gssapi/), [urllib-gssapi](https://github.com/pythongssapi/urllib-gssapi)) that will speak this protocol for you.  I'm happy to provide a patch for either of those if you would prefer.  This package seems to already use urllib2-kerberos (which I have migrated to urllib-gssapi); I'm happy to migrate the remainder of the code to urllib-gssapi as well.

In both cases, you only perform one round of a SPNEGO handshake.  This means that not all SPNEGO connections will complete - you never validate the server tokens, so mutual auth will always fail.  I'm a bit surprised this works at all.  Using requests-gssapi or urllib-gssapi will handle this problem for you.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,7 +8,7 @@ python:
 before_install:
   - "pip install -U pip setuptools virtualenv coveralls"
 install:
-  - "pip install -e .[bitly,kerberos,google]"
+  - "pip install -e .[bitly,gssapi,google]"
   - "git fetch --unshallow"
 script:
   - "coverage run --source=bin,did -m py.test $CAPTURE tests"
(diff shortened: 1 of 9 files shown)


psss:
Thanks for providing the patch. The test suite should now be much better. Currently there does not seem to be `python-urllib2-gssapi` packaged for epel7. Could we use `python2-requests-gssapi` instead?

psss:
I was able to make this working with `requests_gssapi`, but the mutual authentication has to be disabled for the Jira instace I test against. Any ideas for improvement welcome :) Current state merged into master.

PR #149, opened 2018-07-03
Don't zero-fill numbers in Jira issue by default

AloisMahdal:
From my observations, Jira itself does not format issues by including
zeroes, so there's no reason for did to do so.

### DIFF
--- a/did/plugins/jira.py
+++ b/did/plugins/jira.py
@@ -46,7 +46,7 @@ from did.base import Config, ReportError
 from did.stats import Stats, StatsGroup
 
 # Default identifier width
-DEFAULT_WIDTH = 4
+DEFAULT_WIDTH = 0
 
 # Maximum number of results fetched at once
 MAX_RESULTS = 1000


AloisMahdal:
Note that the variable is called `DEFAULT_*` but there does not seem to be way to change it, so maybe the variable could even be removed.  (Maybe in favor of proper way to configure it---if anyone should care?)

psss:
The thing is that leading zeros make the output a bit more readable. We do something similar in the `bugzilla` plugin as well. However there is not consistent approach among plugins. I'm also not sure whether it would make sense to invest time into some user-configurable solution.

AloisMahdal:
> leading zeros make the output a bit more readable

I'd say that's arguable, because if default UI of the ticket system does not zero-fill (most don't), then seeing the zero-filled notation will be actually unfamilliar to the reader.   (Which is why went and created this PR.) This is even emphasized in case of Jira, where the zero-filling happens *inside* the ID:

    OAMG-17           # ikr!
    OAMG-00017        # wt..?

So I agree there's no need to have this configurable by user.  Just show the IDs however the UI shows them.  (Sometimes this might mean that plugin does need extra steps, depending on the API.)

~

Also in general. I think zero-filling makes sense when:

 *  The size of the number in digits is already prescribed by the format specification.  For example, bank
    account numbers, md5 hashes, serial numbers, EAN's...

 *  The resulting strings need to be easily sortable by eg. sort(1) under C locale.  For example,
    the way git format-patch names resulting files.

None of this is the case for ticket id's.

AloisMahdal:
Oh, and another reason for making this consistent with *the ticket system* are [tools like doer][1], that (unlike doer?), will [treat ids][2] in generic way, as strings.  In case of Jira, passing the zero-filled ID back URL actually led to 404, IIRC.

  [1]: https://pagure.io/bmo/
  [2]: https://pagure.io/shellfu-bash-uripecker/blob/master/f/src/uripecker.sh.skel#_6

PR #150, opened 2018-08-23
Fixed problems with Sentry plugin

Madeyro:
However my tests were removed from sentry server so currently there's non of them.


did/plugins/sentry.py
```diff
@@ -112,8 +137,12 @@ def fetch(self):
...
             if (activity['user']['email'] == self.user.email and
                     activity['type'] == 'note'):
-                self.stats.append("{0} - {1}".format(
-                    activity['issue']['shortId'], activity['issue']['title']))
+                record = "{0} - {1}".format(
+                    activity['issue']['shortId'], activity['issue']['title'])
+                # skip if the issue is in the stats already
+                if record in self.stats:
+                    continue
+                self.stats.append(record)
```
psss: What about introducing a common method `SentryStats.append()` which would handle uniqueness instead of copy-pasting the identical code?

did/plugins/sentry.py
```diff
@@ -71,12 +85,19 @@ def filter_data(self):
...
+                if (date > str(self.options.until.date)):
+                    continue
+                if (date < str(self.options.since.date)):
+                    next_page = False
+                    break
+                if (activity['type'] != "set_regression"):
+                    stats.append(activity)
+        # erase cursor, so next stats will be search from the start
+        self.sentry.cursor = ''
         return stats
```
psss: Wouldn't it be better to completely handle the pagination inside `SentryAPI.get_data()`. In this way we could also easily prevent fetching data from the server multiple times.

psss:
> However my tests were removed from sentry server so currently there's non of them.

Could you create a new test content on the server so that we can run CI again?

psss:
> Could you create a new test content on the server so that we can run CI again?

I did that already in c1d648c.

did/plugins/sentry.py
```diff
@@ -51,8 +53,20 @@ def get_data(self):
...
+        if link_header[1].find('results="true"') > 0:
+            # set cursor for next page
+            self.cursor = '?&cursor=' + link_header[1].split('; ')[-1][8:-1]
+        else:
+            self.cursor = ''
         return json.load(response)
 
+    def get_next_page(self):
+        """ Get next page in paginated activity """
+
```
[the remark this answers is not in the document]
Madeyro: I probably wanted to create this new method and then took another approach and forgot to delete it. Sorry for that.

did/plugins/sentry.py
Madeyro: I was happy it works, so not much effort was put into refactoring. Thanks for this review, will do that.

did/plugins/sentry.py
Madeyro: The problem is plugin does not know how much data should pull out of server, so in order to avoid situation when actual data are 5 entries and the plugin would download whole history I did it this way. This could be moved to ```SentryAPI.get_data()``` method by calling ```SentryAPI.get_data(since, until)``` inside ```SentryStats.filter_data()``` which I realize now.  
  
About the fetching data multiple times, I fetch data multiple times because I created the plugin following the bugzilla plugin example. So when you only want to list ```--sentry-resolved``` the plugin would be searching only for resolved.

did/plugins/sentry.py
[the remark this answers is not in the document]
psss: Sure, no problem. So let's just delete it.

did/plugins/sentry.py
psss: Yes, you could limit the search by `since` and `until` and do the download just once. Or, if the API supports more detailed search, then let's use it to prevent fetching of unrelated data. That would be even better than fetching all data for the organization.

For sharing the once-fetched data you can get some inspiration in the `nitrate` plugin:

https://github.com/psss/did/blob/master/did/plugins/nitrate.py#L100

See properties `cases` and `copies` for one way how this could be done. But you could store the data in `SentryAPI` as well. No need to have it under the `StatsGroup` instance.

psss:
@Madeyro, do you plan to look into this? I'm considering a new release in the (hopefully) near future. It would be nice to have this fixed as well.

Madeyro:
> @Madeyro, do you plan to look into this? I'm considering a new release in the (hopefully) near future. It would be nice to have this fixed as well.

Yes, I am planning to fix this. It definitely won't be this weekend but maybe the next one.

psss:
Good. Thanks!

psss:
Thanks for the update. I've squeezed your commits and merged together with a bunch of improvements in 9db27fc (the code did not work at all when there was no paging as test suite revealed).

PR #160, opened 2018-11-30
Support custom plugin locations and arbitrary plugin modules

csams:
Allow users to write their own plugins without needing to download the project repo. They can install `did` and use the config file to point to their custom plugins. This also makes it easier to share plugins outside of the project itself.

* Refactor the plugin loading logic to use [importlib](https://docs.python.org/2/library/importlib.html) and [pkguti](https://docs.python.org/2/library/pkgutil.html)l so it can load plugins from an egg (already on the `PYTHONPATH`) as well as arbitrary places on the file system.
* Add a `custom_plugins` option to the `general` section of the config. Users can add a space or comma separated list of directories where custom plugins reside, and they'll automatically get loaded if the directories are valid python2 packages (include `__init__.py` files).
* Refactor the plugin system so `StatsGroup` subclasses register themselves. This removes the need to scan modules individually and allows plugins to be defined at any level in a package hierarchy.

This PR introduces no new dependencies, but it does use `importlib`, which only comes installed with python 2.7 and beyond. It's available in pypi for python 2.6. The new loader code is python2/3 compatible.

Example config:
```
[general]
email = Joe User <joe@example.com>
custom_plugins = ~/.did/custom_plugins

[thing]
type = thing
```

Example file (thing.py) in ~/did/custom_plugins:
```python
from did.stats import EmptyStatsGroup


class Thing(EmptyStatsGroup):
    """ This is a Thing! """
    pass
```

Example help:
```
└> did -h
usage: did [this|last] [week|month|quarter|year] [options]

optional arguments:
  -h, --help            show this help message and exit

Select:
  --email EMAILS        User email address(es)
  --since SINCE         Start date in the YYYY-MM-DD format
  --until UNTIL         End date in the YYYY-MM-DD format

This is a Thing!:
  --thing               All above

Format:
  --format FORMAT       Output style, possible values: text (default) or wiki
  --width WIDTH         Maximum width of the report output (default: 79)
  --brief               Show brief summary only, do not list individual items
  --verbose             Include more details (like modified git directories)

Utils:
  --config FILE         Use alternate configuration file (default: 'config')
  --total               Append total stats after listing individual users
  --merge               Merge stats of all users into a single report
  --debug               Turn on debugging output, do not catch exceptions
  --test                Run a simple smoke test against the github server
```


csams:
@psss Would you rather this PR go against devel?

psss:
Thanks for the enhancement. This looks good. I just need a bit more time to review the changes. Hopefully in the coming days. No need to update this against the devel branch.

Issue #163, opened 2018-12-03
Jira plugin isn't working anymore with basic auth


angystardust:
Hi, I've just upgraded my _did_ setup to 0.11 (did-0.11-1.fc29.noarch) using the COPR repository but I have to revert back to _did-0.9_ since jira plugin with basic auth is not working anymore.
It seems that now _did_ tries to do a gss auth even if you specify to use the basic auth in the config file:

```
$ did yesterday
Status report for yesterday (2018-12-02 to 2018-12-02).

~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
My User <myuser@example.com>
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 ERROR  401 Client Error:  for url: https://jira.eidosmedia.com/rest/auth/latest/session
 ERROR  Jira authentication failed. Try kinit.
```

My config:

```
[jira-mycompany]
type = jira
prefix = MYGROUP
project = MYGROUP
url = https://jira.example.com/
auth_url = https://jira.example.com/rest/auth/latest/session
auth_type = basic
auth_username = myuser
auth_password = mypassword
```

geoand:
Saw the same exact thing :)

psss:
Thanks for reporting the bug. This should be fixed by 99280c0.

angystardust:
Thank you so much for fixing so fast! :100:

PR #165, opened 2018-12-04
Fix JIRA basic login issue

geoand:
Fixes: #163

### DIFF
--- a/did/plugins/jira.py
+++ b/did/plugins/jira.py
@@ -246,13 +246,11 @@ class JiraStats(StatsGroup):
             self._session = requests.Session()
             log.debug("Connecting to {0}".format(self.auth_url))
             if self.auth_type == 'basic':
-                data = {
-                    "username" : self.auth_username,
-                    "password" : self.auth_password }
                 headers = {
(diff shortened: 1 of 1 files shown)


geoand:
I am not a python expert, so this might not be the best way of doing things, but it at least seemed to work around the JIRA login problem for me

psss:
Thanks for the patch, @geoand. I looked a bit into the requests docs and it seems the solution could be even more straight forward:

    requests.get('https://api.github.com/user', auth=('user', 'pass'))

I've tried to use this approach in 99280c0 and it works for me fine. Could you confirm that it works for you as well? Thanks.

geoand:
You are welcome @psss !

I tried the change you mention and it does indeed work. The only difference I see is that the response in the case of my patch contains a `session` object (which contains things like the `JSESSIONID`). I am note sure if that section are needed or not...

psss:
Great! Thanks for verification. Not sure about the `session` object but created/resolved issues seem to be reported correctly so let's keep it simple.

geoand:
Sounds good :)

Thanks!

PR #166, opened 2018-12-06
Fixing dependencies for Google plugin

theute:
At least this change was required on my machine running on Fedora 29...

### DIFF
--- a/did/plugins/google.py
+++ b/did/plugins/google.py
@@ -41,7 +41,7 @@ by configuring the ``storage`` option.
 import os
 import httplib2
 
-from apiclient import discovery
+from googleapiclient import discovery
 from oauth2client import tools
 from oauth2client.file import Storage
 from oauth2client.client import OAuth2WebServerFlow


psss:
We should probably add `python2-google-api-client` to `Requires`. But it brings a bunch of new dependencies:

    cairo
    cairo-gobject
    fontconfig
    libX11
    libX11-common
    python2-gobject-base
    dejavu-fonts-common
    dejavu-sans-fonts
    fontpackages-filesystem
    libXau
    libXext
    libXrender
    libxcb
    pixman
    python2-SecretStorage
    python2-asn1crypto
    python2-cairo
    python2-cffi
    python2-configparser
    python2-cryptography
    python2-dbus
    python2-entrypoints
    python2-fasteners
    python2-gflags
    python2-httplib2
    python2-keyring
    python2-monotonic
    python2-oauth2client
    python2-ply
    python2-pyOpenSSL
    python2-pyasn1
    python2-pyasn1-modules
    python2-pycparser
    python2-rsa
    python2-uritemplate
    python2-xpyb

Not sure. We could also break `did` into several subpackages to prevent fetching too much dependencies for those who are not interested in all plugins. Or use dnf's weak dependencies? For pip this works very nicely:

    pip install did[google]

Is something like this possible with dnf? Any recommendations?

psss:
That sounds good, @csams. Thanks for the idea. We can do the plugin cleanup/separation later. For now, let's at least update the documentation.

PR #169, opened 2018-12-10
Allow to set 'ssl_verify' config for jira plugin

qiankehan:
Add optional 'ssl_verify' to jira plugin for enabling or disabling ssl
verification. This configuration is true by default.

Signed-off-by: Han Han <hhan@redhat.com>


qiankehan:
Add test function for ssl_verify in jira plugin.

PR #170, opened 2018-12-11
Add commentCard to trello DEFAULT_FILTERS

gfidente:
It is not sufficient to customize the trello plugin 'filters' to
make it display the commentCard objects too, because there is no
mapping for such a card action into a python class.

This change updates the DEFAULT_FILTERS list to include commentCard
and adds the matching filter_map item.


psss:
Thanks for the patch, Giulio. I've fixed the stats title and extended the testsuite in 55326c5 and merged it into master.

gfidente:
thanks a lot!

PR #172, opened 2018-12-14
Adding support for Google tasks

theute:
Note: I never wrote anything in Python, so it's a lot of copy paste...
The patch also includes https://github.com/psss/did/pull/166


theute:
Test fail because the Google user in the tests doesn't have "tasks" permissions.
We should probably check that "tasks" is a listed app in the configuration before trying to fetch those though...

PR #173, opened 2018-12-17
Adding support for Google tasks

theute:
The testsuite will fail, the test account needs additional rights (tasks)


psss:
Thanks for the patch, Thomas. I've added a simple test for completed tasks and updated permissions so tests should now be passing again.

PR #185, opened 2019-01-16
Limit Gerrit data to the specified time

hferreiro:
Previously, the Gerrit plugin would fetch all changes up to the current
date. This commits adds the 'since:' and 'until:' search operators to
limit data fetching to the specified time.

### DIFF
--- a/did/plugins/gerrit.py
+++ b/did/plugins/gerrit.py
@@ -168,6 +168,10 @@ class GerritUnit(Stats):
                 age = (TODAY - self.since_date).days
                 common_query_options += '+-age:{0}d'.format(age)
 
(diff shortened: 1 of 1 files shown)


psss:
Thanks for the patch. Looks good. This should improve the performance. We still need to make some other changes to make it the limit work completely. For example I've noticed that in the reviewed changes there is this strange query:

    reviewer:{0}+is:closed&q=reviewer:{0}+is:open

Which, when extended with since/until results in:

    reviewer:login+is:closed&q=reviewer:login+is:open+since:2019-01-28+until:2019-02-04

But let's do the cleanup later. Going to merge.

PR #186, opened 2019-01-16
Fix some strings in the Gerrit plugin log messages

hferreiro:
This commits fixes wrong CL status in two log messages from the Gerrit
plugin.


psss:
Thanks for the patch!

PR #187, opened 2019-01-16
Add searching for work in progress Gerrit changes

hferreiro:
Add support for searching CLs marked as "Work in progress". These are
CLs for which the code review hasn't started yet. The submitted CLs
report is changed to omit these WIP CLs. Additionally, remove support
for drafts ('is:draft') which isn't available in Gerrit anymore and
probably had the same purpose.


psss:
Looks good. Thanks for the patch. Regarding the `limit_since` parameter: I don't know all the implementation details of the gerrit plugin. It was originally written by @mcepl. Maybe he might have some hint...

PR #188, opened 2019-01-16
Fix the reviewed changes report from Gerrit

hferreiro:
Omit owned CLs from the reviewed changes report. All owned CLs appear as
if the owner is a reviewer too, but it doesn't make sense to include
them in this report.


psss:
Makes sense. Thanks for the patch. I just had to update corresponding test because the example result was among those which are now filtered out.

PR #192, opened 2019-03-14
gerrit: also include the project name

tosky:
It is useful to see the project (repository) of each review
if you contribute to multiple projects on the same gerrit server.

The drawback is that the commit message may be even shorter.


psss:
Looks good. We can add some config option later if needed. Thanks for the patch.

PR #194, opened 2019-03-14
gerrit: add a new configuration key to disable wip queries

### LINKED ISSUE #193: gerrit plugin fail to list changes when the server does not support the wip flag
A recent change (689c6a57c9c7e01b60ea41fe32bf3ce6b8258968, post-0.12, not yet released) implemented the support for is:wip flag, removing is:draft. Unfortunately not all servers support it, because they didn't upgrade yet. Notably, review.openstack.org and review.rdoproject.org.

It should be possible to configure the server in a way that ignores the wip flag.

tosky:
Not all gerrit servers support the is:wip search. Trying to search
for it on such servers lead to a 400 error.
A new flag can be added to each server to disable wip-based searches.
The default behavior is unchanged, as hopefully more and more
server will support this feature.
The code provides the groundwork to support more of such feature flags.

Fixes #193.

### DIFF
--- a/did/plugins/gerrit.py
+++ b/did/plugins/gerrit.py
@@ -8,6 +8,9 @@ Config example::
     type = gerrit
     url = https://example.org/gerrit/
     prefix = GR
+    # optional, True by default; set to False if the gerrit server
+    # does not support wip as search criteria.
+    wip = True
 """
 
 import json
@@ -125,6 +128,7 @@ class GerritUnit(Stats):
         self.prefix = prefix if prefix is not None else parent.config['prefix']
         self.repo = Gerrit(baseurl=self.base_url, prefix=self.prefix)
         self.since_date = None
+        self.server_features = [] if not parent else parent.server_features
 
         Stats.__init__(self, option, name, parent)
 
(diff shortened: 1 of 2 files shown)


psss:
Looks good. Could you add at least a short mention about the flag in the module docs? Thanks.

tosky:
I hope it's ok - I didn't find a cleaner way to document the options.

tosky:
Ping - should I increase the coverage (even though the decrease is really negligible)?

psss:
Sorry for late response. Yes, it would be good to include a simple test with the `wip` disabled. Also one more thought: Instead of propagating the server features into each stats we could just store the value in the parent stats and reference them e.g. `self.parent.features`. This is quite often used in other plugins as well.

tosky:
Ok, let's see if it works now.

tosky:
The tests failed, but the only failure seems to be unrelated to this change; the bugzilla test failed.

tosky:
ping :)

tosky:
Would it be worth to re-execute the tests after the last optimizations?

psss:
Not needed. I've tested the changes on the devel branch. Everything looks fine. Merging. And sorry for my late response.

PR #195, opened 2019-06-05
Improved documentation of redmine - DB ID

lzap:
Spent some time debugging why redmine did not work until I realized it's database ID not user login name. Made the text more clear about it to prevent confusion.


psss:
Thanks for improving the description!

PR #197, opened 2019-06-10
Add last friday command

lzap:
I report on a daily basis and I miss ability to report last Friday. This should do it, also support next Friday which does not make sense, but it's good to be consistent. Not a pythonist, so take care reviewing.

I wanted to add a more generic "last business day" command but it turns out that it's tough to do business days in plain Python and I do not want to add another dependency because of this. If you want the patch to be Israel-friendly, then I can add also "last sunday" or maybe all the days but that could be little bit noisy.

### DIFF
--- a/did/base.py
+++ b/did/base.py
@@ -296,6 +296,17 @@ class Date(object):
             until = Date("yesterday")
             until.date += delta(days=1)
             period = "yesterday"
+        elif "friday" in argument:
+            since = Date("today")
+            until = Date("today")
+            if "last" in argument:
+                since.date += delta(weekday=FR)
+                until.date += delta(weekday=FR)
+                period = "last friday"
+            else:
+                since.date += delta(weekday=FR(-1))
+                until.date += delta(weekday=FR(-1))
+                period = "next friday"
         elif "year" in argument:
             if "last" in argument:
                 since, until = Date.last_year()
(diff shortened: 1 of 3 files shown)


AloisMahdal:
BTW, you could hack around it by something like:

    did --since `date -d 'last friday' +%Y-%m-%d` --until `date -d 'last friday' +%Y-%m-%d`

or

    did_1day() {
        #
        # Report activity for day $1 (in format accepted by date(1) -d)
        #
        local expr=("$@")
        local day
        day=$(date -d "${expr[*]}" +%Y-%m-%d) || {
            echo "could not convert date: '${expr[*]}'" >&2
            return 3
        }
        did --since "$day" --until "$day"
    }

and then just call

    did_1day last friday

..in the end it would be much more poweful than adding every expression as extra command, since it can do many other tricks.

Or, maybe did could have a way to convert the expression internally?

lzap:
Hmm I did not see "NameError: global name 'FR' is not defined" locally, this is probably defined in the delta module. Anyway, let me know if you want this patch otherwise I'd just drop it given the workaround (nice thanks).

psss:
Thanks for the patch. Looks fine. Going to merge after a couple of adjustments in 06bae06. We can extend date selection to be more generic in the future. The `last friday` seems that it could be a relatively common use case so it's worth to include it now.

PR #198, opened 2019-07-11
Use scriptrunner issueFunction to speed up things

sochotnicky:
Original code needed per-project JIRA definition because JiraUpdated was
querying all JIRA issues in a project that changed over defined period
of time. This was wasteful and for JIRA instances with scriptrunner
installed we can do things faster and with a single configuration

With this change:
 * project configuration is optional - if omitted all projects are
   reported
 * added new "login" config item needed because when lastUpdated('by
   <login>') is triggered with non-existing login name it causes failure
   of the JIRA plugin.

This will need further rework so that use of scriptrunner is
optional/configured properly (since it's not part of JIRA itself). Obviously docs need updates too. I'll try to clean this up later

### DIFF
--- a/did/plugins/jira.py
+++ b/did/plugins/jira.py
@@ -7,23 +7,35 @@ Configuration example (GSS authentication)::
     [jboss]
     type = jira
     prefix = JIRA
-    project = ORG
     url = https://issues.jboss.org/
     ssl_verify = true
 
-Configuration example (basic authentication)::
+Configuration example (basic authentication) with alternative username::
 
     [jboss]
     type = jira
     prefix = JIRA
-    project = ORG
+    login = alt_username
     url = https://issues.jboss.org/
     auth_url = https://issues.jboss.org/rest/auth/latest/session
     auth_type = basic
     auth_username = username
     auth_password = password
 
+Configuration example limiting report only to a single project::
+
+    [jboss]
+    type = jira
+    prefix = JIRA
+    project = ORG
+    url = https://issues.jboss.org/
+    ssl_verify = true
+
 Notes:
+* If your JIRA does not have scriptrunner installed you must set
+  ``use_scriptrunner`` to false
+* You must provide ``login`` variable that matches username if it
+  doesn't match email/JIRA account
 * Optional parameter ``ssl_verify`` can be used to enable/disable
   SSL verification (default: true)
 * ``auth_url`` parameter is optional. If not provided,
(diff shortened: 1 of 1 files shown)


sochotnicky:
Also of note might be that - JIRA 8.x will support this query out of the box it seems:
https://jira.atlassian.com/browse/JRASERVER-1973

psss:
Thanks for the patch. It would be nice to be able to easily search across all projects.

sochotnicky:
Now the plugin should work with or without scriptrunner though we try to use it by default since it's a fairly common plugin (and speeds up execution)

psss:
I tried to run a Jira report against our internal instance with my current config but it did not work. Disabling the script runner using `use_scriptrunner = false` helped. I think it would not be good to break reports for current users with their configs already set. Could you switch the default method, so that script runner is enabled only when `use_scriptrunner = true` is present in the config?

sochotnicky:
> I tried to run a Jira report against our internal instance with my current config but it did not work. Disabling the script runner using `use_scriptrunner = false` helped. I think it would not be good to break reports for current users with their configs already set. Could you switch the default method, so that script runner is enabled only when `use_scriptrunner = true` is present in the config?

Yeah, I left out a call to Issue.search during some rebase/merge. Hopefully that should be enough to fix it. If not I can change the defaults but fingers crossed...

psss:
We're almost there ;-)

```
Status report for the week 15 (2020-04-06 to 2020-04-12).

~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 Petr Šplíchal <psplicha@redhat.com>
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

* Issues created in tt: 2
    * TT-105 - Testing Tools Retrospective
    * TT-104 - Update Fedora CI documentation with tmt instructions and...
  File "/home/psss/.virtualenvs/did/bin/did", line 7, in <module>
    exec(compile(f.read(), __file__, 'exec'))
  File "/home/psss/git/did/bin/did", line 42, in <module>
    did.cli.main()
  File "/home/psss/git/did/did/cli.py", line 222, in main
    user_stats.check()
  File "/home/psss/git/did/did/stats.py", line 157, in check
    stat.check()
  File "/home/psss/git/did/did/stats.py", line 157, in check
    stat.check()
  File "/home/psss/git/did/did/stats.py", line 77, in check
    self.fetch()
  File "/home/psss/git/did/did/plugins/jira.py", line 177, in fetch
    self.stats = Issue.search(query, stats=self)
  File "/home/psss/git/did/did/plugins/jira.py", line 120, in search
    batch, listed(data["issues"], "issue")))
KeyError: 'issues'
```

psss:
Thanks, seems to be working fine now. Merged into master.

PR #199, opened 2019-07-12
Add basic Confluence support

sochotnicky:
This only supports creation of comments/pages and basic JIRA code since they share a lot of things around auth. We could add blogposts too if there's need

I did not find a way to query modifications by a specific person so that's missing unfortunately


psss:
Thanks for writing the plugin, Stano. Looks good. There was only some problem with fetching again and again the same pages from the server. I've fixed that and with a couple of other adjustments in c7c2c69 merged.

PR #209, opened 2019-11-12
Google plugin __unicode__ leftover

lukaszachy:
Changed to __str__ as the rest


psss:
Thanks for catching this, Lukáš.

PR #215, opened 2019-12-09
Enable simple smoke test in the testing farm

psss:


.packit.yaml
```diff
@@ -21,3 +21,12 @@ jobs:
     - fedora-31-x86_64
     - fedora-rawhide-x86_64
   trigger: pull_request
+
+jobs:
```
TomasTomecek:
```suggestion
```
with this, you redefine jobs and thus remove copr-build :P no, packit is not taking a nap

psss: Oh! I see. Bad copy and paste. Thanks for spotting this.

jpopelka:
just testing @packit-service/the-packit-team

PR #218, opened 2020-01-06
Add default command into the info log [fix #217]

### LINKED ISSUE #217: Getting all the switches used by default
When running the git like "did last" having multiple services configured, I would like to see the an equivalent command line with all the service specific switches included. It would let me to easily fine tune the parameters instead of having to reconstruct it on my own.
The output doesn't have to be covered by and extra switch, I would be fine with seeing the output e.g. in the verboseverbose (vv) or debug log. Or it could be some git config subcommand, all are really care about are the switches on by default.

Example:
$ did last
having and equivalent
$ did --bz --jira --nitrate --wiki last

psss:
Resolves #217.

### DIFF
--- a/did/cli.py
+++ b/did/cli.py
@@ -70,6 +70,8 @@ class Options(object):
         log.debug("Loading Sample Stats group to build Options")
         self.sample_stats = UserStats()
         self.sample_stats.add_option(self.parser)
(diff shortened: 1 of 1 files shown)


psss:
Looks good, just irrelevant errors from other plugins.

PR #220, opened 2020-02-17
Do not break sorting on merge

sochotnicky:
My original code for deduplicating in db503e15 was naive and causes another
minor problem when ordering of reported items get reshuffled by set().

This change makes it a little more conservative and when merging we iterate over
list and only append when merging. This way ordering is preserved better and not
randomized.

### DIFF
--- a/did/stats.py
+++ b/did/stats.py
@@ -101,9 +101,9 @@ class Stats(object):
 
     def merge(self, other):
         """ Merge another stats. """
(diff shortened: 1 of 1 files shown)


psss:
Thanks for improving the deduplication. I tried it on Jira tickets but it did not work because there was missing a comparison method. Thanks to 71e4491 this should now be working for Jira issues as well.

PR #225, opened 2020-07-09
Enable epel-8 for packit testing

psss:


psss:
It looks like `python3-httplib2` is missing in `epel8`. Filed [BZ#1856282](https://bugzilla.redhat.com/show_bug.cgi?id=1856282]) to track this.

psss:
The bug has been closed as wontfix because `python3-httplib2` is available in PowerTools. @thrix, seems we should enable PowerTools in cruncher.

psss:
@thrix, what do you think about enabling the PowerTools repo in cruncher? See also above.

PR #229, opened 2020-08-08
Add support for loading Jira/Confluence password from a file

### LINKED ISSUE #228: Add support for loading Jira/Confluence password from a file
When I use basic authentication for Jira or Confluence, the `auth_password` option must be set in `~/.did/config`.

However, saving passwords directly to the configuration file is not so great.
When we currently use `auth_password`, we mix configuration parts and passwords, and we also must to fill a password in each Jira/Confluence section.
I think it would be good enhancement if we can add support for loading passwords from separate valut/password files.

We need to store the path to a vault file somewhere.
So we need to add a new option. Something like: `auth_valut`, `auth_password_valut`, `auth_valut_file` or `auth_password_file`.

examples of such a configuration
auth_username = user
auth_valut_file = ~/.did/.valut_jira

auth_username = user
auth_valut_file = ~/.did/.valut_confluence

jharuda:
resolves #228

### DIFF
--- a/did.spec
+++ b/did.spec
@@ -1,5 +1,5 @@
 Name: did
-Version: 0.16
+Version: 0.17
 Release: 1%{?dist}
 
 Summary: What did you do last week, month, year?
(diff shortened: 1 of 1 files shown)


psss:
Sorry, pushed wrong commit by mistake. Moved to #230.

PR #233, opened 2020-11-03
base.py: Use none interpolation for ConfigParser

qiankehan:
When character '%' is used in auth_password, the ConfigParser will raise
InterpolationSyntaxError because the ConfigParser in default interpolation[1]
parses '%' as formatting character.

It will report the error like:
configparser.InterpolationSyntaxError: '%' must be followed by '%' or '(', found: '%'

[1]: https://docs.python.org/3.6/library/configparser.html#interpolation-of-values

Signed-off-by: Han Han <hhan@redhat.com>

### DIFF
--- a/did/base.py
+++ b/did/base.py
@@ -84,7 +84,7 @@ class Config(object):
         # Read the config only once (unless explicitly provided)
         if self.parser is not None and config is None and path is None:
             return
-        Config.parser = configparser.ConfigParser()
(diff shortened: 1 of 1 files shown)


psss:
Looks good. Thanks for the improvement! For future pull requests, please consider writing the main [commit message](https://did.readthedocs.io/en/latest/contribute/#commits) as a full sentence without any prefix. Thanks.

PR #236, opened 2021-02-08
plugins/{git,google}: expand tilde in config path values

stepnem:

PR #237, opened 2021-02-09
Add "no prefix" rule to commit message suggestions

stepnem:
Cf.
https://github.com/psss/did/pull/236#issuecomment-775759894-permalink


psss:
Nice, thanks!

PR #251, opened 2021-10-18
Add basic support for Zammad

come-nc:
https://zammad.org/


did/plugins/zammad.py
```diff
@@ -0,0 +1,121 @@
...
+        try:
+            request = urllib.request.Request(url, headers=self.headers)
+            response = urllib.request.urlopen(request)
+            log.debug("Response headers:\n{0}".format(
+                str(response.info()).strip()))
+        except urllib.error.URLError as error:
+            log.debug(error)
+            raise ReportError(
+                "Zammad search on {0} failed.".format(self.url))
+        result = json.loads(response.read())["assets"]["Ticket"]
```
lukaszachy, line 54:
```suggestion
        try:
            result = json.loads(response.read())["assets"]["Ticket"]
        except KeyError:
            result =  dict() 
```
Problem is that if  API changes this will happily return empty results.

lukaszachy: Hi @come-nc. Another option is first do `result = json.loads(response.read())["assets"]` and after that try/except `result = result['Ticket']`

come-nc: Thank you, it works.

psss (approved):
Looks good, would be nice to have some public testing instance to verify the functionality, but we can add that later.

psss:
Added the new plugin to the `plugins.rst` file so that the new config appears in the documentation as well, rebased on the latest master, squashed and merged. Thanks!

Issue #254, opened 2021-10-31
Bugzilla plugin does not paginate or have limit option


mikelolasagasti:
Last september it was announced that RedHat's bugzilla will be update to include pagination. By default all results will be limited to 20 results.

https://communityblog.fedoraproject.org/changes-to-bugzilla-queries/

I can't find how to change this in `did`, so I assume is not implemented.

I verified this with `bugzilla` command:

Create login:
```$ bugzilla login --api```

Launch bugzilla command without limit:
```
$ bugzilla --bugzilla https://bugzilla.redhat.com/bugzilla/xmlrpc.cgi --debug query --field=f1=reporter --field=o1=equals --field=v1=mikel@olasagasti.info --field=query_format=advanced
[12:20:55] DEBUG (_cli:1238) Launched with command line: /home/mikel/Proiektuak/did/env/bin/bugzilla --bugzilla https://bugzilla.redhat.com/bugzilla/xmlrpc.cgi --debug query --field=f1=reporter --field=o1=equals --field=v1=mikel@olasagasti.info --field=query_format=advanced
[12:20:55] DEBUG (_cli:1239) Bugzilla module: <module 'bugzilla' from '/home/mikel/Proiektuak/did/env/lib64/python3.10/site-packages/bugzilla/__init__.py'>
[12:20:55] DEBUG (_authfiles:211) Using cookiefile=/home/mikel/.cache/python-bugzilla/bugzillacookies
[12:20:55] DEBUG (_authfiles:181) Using tokenfile=/home/mikel/.cache/python-bugzilla/bugzillatoken
[12:20:55] INFO (_authfiles:67) Found bugzillarc files: ['/home/mikel/.config/python-bugzilla/bugzillarc']
[12:20:55] DEBUG (base:513) Connecting with URL https://bugzilla.redhat.com/bugzilla/xmlrpc.cgi
[12:20:55] DEBUG (_authfiles:89) bugzillarc: Searching for config section matching https://bugzilla.redhat.com/bugzilla/xmlrpc.cgi
[12:20:55] DEBUG (_authfiles:101) bugzillarc: Found matching section: bugzilla.redhat.com
[12:20:55] DEBUG (base:438) bugzillarc: setting api_key
[12:20:55] DEBUG (base:532) using API key
[12:20:55] DEBUG (_backendxmlrpc:120) XMLRPC call: Bugzilla.version({})
[12:20:57] DEBUG (base:535) Bugzilla version string: 5.0.4.rh64
[...]

PR #257, opened 2021-11-15
Return all the bugzilla results from a search

jbastian:
As of 2021-09-13, bugzilla.redhat.com only returns 20 results (by
default) from a search.  The limit parameter can override this (up to
1000 results).  A limit of 0 is the same as max, so use 0 in did.

https://bugzilla.redhat.com/docs/en/html/integrating/api/Bugzilla/WebService/Bug.html?highlight=limit#search

Resolves issue #254


psss (approved):
Thanks for the fix!

PR #259, opened 2021-12-01
Support Personal Access Token in the Jira plugin

lukaszachy:


lukaszachy:
@psss  Feel free to squash commits into one. It works for me locally (tm)

psss (approved):
Thanks a lot for implementing the token authentication. I'm proposing a couple of improvements in 996d120. Basically reorganized the config examples to make them a bit more readable as the section has grown substantially and the long original notes at the end were hard to read.

Maybe one question: When looking at the config for the first time I was not able to parse the `days:token-name` format. So I'm thinking whether, perhaps, something like the following example could be more readable at the first sight?

```ini
[jira]
type = jira
url = https://issues.redhat.com/
auth_type = token
token_file = ~/.did/jira-token
token_expiration = 33
token_name = did-token
```

What do you say? Would it be worth it?

lukaszachy:
> What do you say? Would it be worth it?

Yes, I'll update PR. Thing is that token_name will be updated quite often and it will be easier to edit.

psss:
Nice, looks much better now! Thanks for the improvement. Merged.

PR #260, opened 2021-12-02
Add minor clarification note about token name

AloisMahdal:
I did not get at the first time that the name is the actual token name,
not just some local alias.

On the second thought, it seems kina obvious, but perhaps the note might
still help someone.


psss (approved):
Makes sense, let's be as explicit and precise as possible ;-) Thanks for the improvement!

PR #262, opened 2021-12-04
Check if the GitHub credential token is valid

mikelolasagasti:
GitHub return 401 when configured credential token is not valid. As the
token is optional, show a message requesting either to update the token
or just to remove it.

### DIFF
--- a/did/plugins/github.py
+++ b/did/plugins/github.py
@@ -67,6 +67,13 @@ class GitHub(object):
                 log.debug(error)
                 raise ReportError(f"GitHub search on {self.url} failed.")
 
+            # Check if credentials are valid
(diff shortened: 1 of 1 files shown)


psss (approved):
Thanks for the improvement! Looks good. Going to merge after a minor style adjustment.

psss:
@mikelolasagasti, note that we are not using prefixes in commits, labels are used for categorization instead. You might want to have a look at other commit [recommendations](https://did.readthedocs.io/en/latest/contribute.html#commits) as well.

mikelolasagasti:
Thanks @psss I'll check for next time. As some other projects like `sos` require prefix for each plugin, I didn't thought twice and didn't check project's style.

psss:
Sure, no problemo :)

PR #271, opened 2022-08-08
handle GitHub API rate limit

hrw:
GitHub API has serious rate limit for tokenless queries.

So instead of failing with traceback just print info that GitHub API
rate limit was exceeded.


psss (approved):
Thanks for improving this! Looks good. I've only adjusted the error message a bit.

psss:
Thanks for the fix @jpopelka, now everything's green.

PR #277, opened 2022-10-24
feat: token_file config support for github plugin

kwk:
This adds the possibility to not only use the `token_file` setting with
the github plugin. The overall goal is to make did-config files more
sharable and when you have the `token` visible within your config, this
is potentially more dangerous.


kwk:
superseded by #278

PR #278, opened 2022-10-24
Allow all plugins to fetch secrets from files

kwk:
Some plugins like the Jira one allowed to store their token used for
authentication within a file and have you point to the file using the
`token_file` config option. I liked that idea because it makes did
config files much more shareable amongst team members.

That's why I've added the token-in-file mechanism to all plugins. The
google plugin also stores a `client_it` and `client_secret` that you can
also put into a file now by specifying `client_id_file` and
`client_secret_file`.

I've added tests for the `get_token` function and for the github plugin
to utilize this function when running in a Github action because then
the environment variable `GITHUB_TOKEN` is set and can be tested when
putting it in a token file only for it to be consumed then.

### DIFF
--- a/did/base.py
+++ b/did/base.py
@@ -457,3 +457,45 @@ class User(object):
         if login is not None:
             self.login = login
             log.info("Using login alias '{0}' for '{1}'".format(login, stats))
+
+
+def get_token(
+        config: dict,
+        token_key: str = "token",
+        token_file_key: str = "token_file") -> str:
+    """
+    Extract the authentication token from config or token file
+
+    Returns the contents of `config[token_key]`, or the file contents of
+    `config[token_file_key]` if no `config[token]` exists. If neither
+    keys exist, `None` is returned.
+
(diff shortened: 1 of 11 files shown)


psss (approved):
Thanks for implementing this improvement. Very nice patch! Added just a couple of minor style adjustments in c5e077c.

kwk:
@psss in order to fix the error with copr, you'd have to merge this PR first: https://github.com/psss/did/pull/285

kwk:
@psss I've rebased this patch after #284 landed and the errors with the man page are gone now. I would appreciate if you can give it another review and merged the PR. Please squash the commits when merging as they are mostly fix-up commits. I suggest to use the PR description as commit message.

psss:
Great, thanks! Squashed and rebased.

kwk:
> Great, thanks! Squashed and rebased.

Oh, I thought you'd squash when you merge in one go. I'm doing this with the drop down in the github merge button. But it depends on your workflow I guess. Thank you.

PR #281, opened 2022-11-12
Add configuration options for border

mweetman-redhat:
This change allows you to change the border char, or have no border with `border_char=` in your config. It also allows you to change the width of the border.


sandrobonazzola (approved):
LGTM

did/base.py
```diff
@@ -150,6 +150,22 @@ def width(self):
         except (NoOptionError, NoSectionError):
             return MAX_WIDTH
 
+    @property
+    def border_char(self):
+        """ Border character to use for the report """
```
psss, line 155: I'm not sure about the naming as `border` confused me a bit when reading the pull request. What about `separator`? Does that sound ok? @lukaszachy, any better idea?

did/base.py
```diff
@@ -150,6 +150,22 @@ def width(self):
...
+        try:
+            return self.parser.get("general", "border_char")
+        except (NoOptionError, NoSectionError):
+            return "~"
+
+    @property
+    def border_count(self):
```
psss, line 162: What about syncing the separator count with the report `width`? It would nicely go together when the line width is increased. Except for `width=0` which is used for unlimited line width.

did/cli.py
```diff
@@ -218,7 +218,7 @@ def main(arguments=None):
             if options.merge:
                 utils.item(user, 1, options=options)
             else:
-                utils.header(user)
+                utils.header(user, did.base.Config().border_count, did.base.Config().border_char)
```
psss, line 221: Shouldn't the same format be used for the `Total Report` header as well?

psss (changes requested):
Thanks for the improvement! Looks nice. Added a few comments. Also, could you please extend the [general section](https://did.readthedocs.io/en/latest/config.html#general) of the config documentation so that it mentions the newly added key(s)? Thanks.

did/base.py
mweetman-redhat: I implemented these changes for a friend who specifically didn't want it to match the screen width, so I'd have to disagree with this one.

did/base.py
mweetman-redhat: Happy to change the name to whatever you prefer

did/cli.py
mweetman-redhat: Added

did/base.py
mweetman-redhat: I won't update the docs until everything is finalised

did/base.py
psss: Ok, I see, understood.

did/cli.py
psss: There's actually one more instance of the `header()` on line 213.

did/base.py
psss: Ok, let's use `separator` instead of `border_char` and `separator_width` for `border_count`. Thanks.

did/cli.py
mweetman-redhat: Got it

did/base.py
mweetman-redhat: Done

mweetman-redhat: Doco updated

mweetman-redhat: I was concerned about renaming `border_char` to `separator` as it might cause confusion with other uses, for example: `emails = utils.split(emails, separator=re.compile(r"\s*,\s*"))` .. wdyt?

mweetman-redhat: Done

mweetman-redhat:
Hmm... build failed, are you sure we aren't conflicting using `separator` as an option?

psss:
The `pre-commit` failure is unrelated. I've updated the `isort` version which should fix the problem.

psss:
Seems that GitHub does not allow to reopen pull requests created from the `main` branch. Created a new one.

PR #284, opened 2023-01-20
Update packit config to address recent changes

psss:
The `srpm_build_deps` section is now mandatory.
The copr build should be specified explicitly.

### DIFF
--- a/.packit.yaml
+++ b/.packit.yaml
@@ -4,21 +4,36 @@ actions:
   get-current-version:
   - make version
 
+srpm_build_deps:
+  - make
+  - python3-docutils
+
 jobs:
+- job: copr_build
+  trigger: pull_request
+  targets:
+  - fedora-all
+  - epel-8
+  - epel-9
+  enable_net: False
+
 - job: tests
   trigger: pull_request
   metadata:
     targets:
     - fedora-all
     - epel-8
+    - epel-9
 
 - job: copr_build
   trigger: commit
   metadata:
-    branch: master
+    branch: main
     targets:
     - fedora-all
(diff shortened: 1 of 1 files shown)


kwk:
I think this is more than what is actually needed. The minimal change to fix the Copr issue is here: https://github.com/psss/did/pull/285

psss:
Thanks! It seems we've created the pull requests at the very same minute :) Actually there were some more adjustments which should be done for the `packit` configuration (copr build detection from testing jobs will be removed in the future), so I've included them all here. The `python3-docutils` requirement, however, seems to be unnecessary. Will remove it.

lukaszachy:
> Thanks! It seems we've created the pull requests at the very same minute :) Actually there were some more adjustments which should be done for the `packit` configuration (copr build detection from testing jobs will be removed in the future), so I've included them all here. The `python3-docutils` requirement, however, seems to be unnecessary. Will remove it.

Are you sure? SRPM cannot be built without `python3-docutils` (it is used during `make srpm` call) . Depends on what is installed by default however on reasonably minimal installs it was missing.

psss:
> Are you sure? SRPM cannot be built without `python3-docutils`...

Yes, that was my impression as well. And I remember adding this one to the config when setting up `tmt`. But, apparently, the build in #285 succesed. Ha! But the `man` page was not created correctly:

```
:: [ 10:16:33 ] :: [   FAIL   ] :: File 'man' should contain 'What did you do last week, month, year?' 
```

So I'll keep it there.

kwk:
LGTM

PR #286, opened 2023-01-20
Update `pre-commit` configuration, enable `flake8`

psss:
Add `flake8` to the list of enabled linters.
Update the maximum line width settings.
Fix newly discovered issues.

### DIFF
--- a/did/plugins/trac.py
+++ b/did/plugins/trac.py
@@ -96,7 +96,9 @@ class Trac(object):
             for ticket, changelg in zip(tickets, changelogs)]
 
     def history(self, user=None):
-        """ Return relevant who-did-what logs from the ticket history """
+        """
+        Return relevant who-did-what logs from the ticket history
+        """
         for event in self.changelog:
             when, who, what, old, new, ignore = event
             if (when >= self.options.since.date and
@@ -112,7 +114,9 @@ class Trac(object):
         return False
 
     def updated(self, user):
-        """ True if the user commented the ticket in given time frame """
+        """
+        True if the user commented the ticket in given time frame
+        """
         for who, what, old, new in self.history(user):
             if (what == "comment" or what == "description") and new != "":
                 return True
@@ -165,8 +169,8 @@ class TracUpdated(TracCommon):
 
     def fetch(self):
         log.info("Searching for tickets updated by {0}".format(self.user))
-        query = "time=..{2}&modified={1}..".format(
-            self.user.login, self.options.since, self.options.until)
+        query = "time=..{1}&modified={0}..".format(
+            self.options.since, self.options.until)
         self.stats = [
             ticket for ticket in Trac.search(query, self.parent, self.options)
             if ticket.updated(self.user)]
--- a/did/utils.py
+++ b/did/utils.py
@@ -9,7 +9,7 @@ import pkgutil
 import re
 import sys
 import unicodedata
-from pprint import pformat as pretty
+from pprint import pformat as pretty  # noqa: F401 (used by other modules)
 
 # ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 #  Constants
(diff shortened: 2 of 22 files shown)


tests/plugins/test_redmine.py
```diff
@@ -1,12 +1,8 @@
...
 """ Tests for the Redmine plugin """
 
-import time
-
-import pytest
-
-import did.base
-import did.cli
+# import did.base
+# import did.cli
```
kwk, line 5: Why leave the commented out imports in here?

docs/contribute.rst
```diff
@@ -35,6 +35,7 @@ It is challenging to be both concise and descriptive, but that is
 what a well-written summary should do. Consider the commit message
 as something that will be pasted into release notes:
 
+* Maximum line length is 88 for code and 72 for docs.
```
kwk, line 38: Why not make both 88?

docs/conf.py
```diff
@@ -167,62 +167,62 @@
...
 
 # If true, an OpenSearch description file will be output, and all pages will
 # contain a <link> tag referring to it.  The value of this option must be the
 # base URL from which the finished HTML is served.
-#html_use_opensearch = ''
+# html_use_opensearch = ''
 
 # This is the file name suffix for HTML files (e.g. ".xhtml").
-#html_file_suffix = None
+# html_file_suffix = None
```
kwk, line 211: What's the matter with all these spaces after the comment? Does the space matter?

did/utils.py
```diff
@@ -9,7 +9,6 @@
 import re
 import sys
 import unicodedata
-from pprint import pformat as pretty
```
kwk, line 12: Oh, this can have a bad consequence as some library or file down the road will be missing `pretty`. I've remove this line once and moved it back in because I was too lazy to fix it properly.

did/plugins/trac.py
```diff
@@ -165,8 +165,8 @@ class TracUpdated(TracCommon):
 
     def fetch(self):
         log.info("Searching for tickets updated by {0}".format(self.user))
-        query = "time=..{2}&modified={1}..".format(
-            self.user.login, self.options.since, self.options.until)
+        query = "time=..{1}&modified={0}..".format(
+            self.options.since, self.options.until)
```
kwk, line 169: Now that you're at it, why not go with the following?

```Python
query = f"time..{self.options.until}&modified={self.options.since}.."
```

How did the 2-1 thing get in here in the first place? Isn't there a test to check this?

kwk (commented):
I have some questions/concerns...

did/utils.py
psss: Good point! Actually, `pretty` is used by other modules. Fixed in 82dd820. Thanks!

kwk: You're very welcome!

tests/plugins/test_redmine.py
psss: Because of the disabled test at the bottom of the file. When the time comes to resurrect it, everything's there.

docs/contribute.rst
psss: Because that would go against the [pep8 recommendations](https://peps.python.org/pep-0008/#maximum-line-length). And I agree with them that from the readability point of view this is a good maximum.

docs/conf.py
psss: Just to make `autopep8` happy. There's a recommendation to always add a space of the `#` sign. And it doesn't hurt here.

did/plugins/trac.py
psss: We could convert to format strings but  `format()` is used over 20 times across the file so let's keep that for a separate pull request to keep the code consistent.

> How did the 2-1 thing get in here in the first place? Isn't there a test to check this?

Well,, probably just a copy & paste of the other stat. Trac is not used much recently so I guess the plugin will eventually die. Test improvements are always welcome ;)

docs/conf.py
kwk: I'm just worried about the noise this PR adds.

did/plugins/trac.py
kwk: Haha, not from me ;)

docs/conf.py
psss: Understood, but that's life ;-)

PR #287, opened 2023-01-23
Add "name-tests-test" pre-commit hook

kwk:
The `name-tests-test` hook verifies that test files are named "correctly". That's why all test files have to start with `test_*.py`.

See https://pre-commit.com/hooks.html

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -45,3 +45,7 @@ repos:
       - id: end-of-file-fixer
       - id: mixed-line-ending
       - id: trailing-whitespace
+      - id: name-tests-test
+        args:
+          - --pytest-test-first
+        exclude: disable_test_trac.py


psss:
What's the motivation for moving the `test` prefix to suffix? Is there any benefit for that?

kwk:
> What's the motivation for moving the `test` prefix to suffix? Is there any benefit for that?

Well, I'd say that the benefit is, that we can check for test file naming this way. This documentation shows that you can configure the hook with options to have the check do a prefix test for `test_*.py`: https://github.com/pre-commit/pre-commit-hooks#name-tests-test.I have enabled that and then we can leave alone the renaming. The only exclusion is `disable_test_trac.py`. Would this be fine?

psss (approved):
Looks good, thanks.

PR #289, opened 2023-01-27
Fix broken phabricator tests

kwk:
Fixes:

* I've wrongly used format strings before.
* Also, this test fixed the `did.base.OptionError: Invalid argument: '--verbose--since'` error that was showing without the additional spaces added by this patch.
* Fix `test_missing_token`: The `CONFIG_BAD_MISSING_TOKEN` string was identical to the `CONFIG_OK` string and thus didn't raise `ConfigError` as expected when the token part is missing.

### DIFF
--- a/tests/plugins/test_phabricator.py
+++ b/tests/plugins/test_phabricator.py
@@ -25,7 +25,7 @@ type = phabricator
 
 URL = "url = https://reviews.llvm.org/api/"
 LOGINS = "login = kwk, kkleine"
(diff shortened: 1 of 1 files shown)


psss (approved):
Looks good, thanks! As for the commit summary, please have a look at [recommendations](https://did.readthedocs.io/en/latest/contribute.html#commits). Let's rather use github labels for categorizing issues and pull requests. Thanks.

Issue #288, opened 2023-01-26
phabricator plugin shows reviews in which reviewer hasn't done any work yet


kwk:
My colleage @nikic noticed that the `phabricator` plugin lists reviews in which he is just a reviewer but hasn't done any work yet. This is of course not our goal. Such reviews show up under `Reviews participated on <SECITON_NAME>`.

PR #290, opened 2023-01-27
[phabricator] unified, pageable request methods

kwk:
This is done in order to prepare for more calls that we have to make to
phabricator and to not repeat the request logic all the time.

See https://github.com/psss/did/issues/288


kwk:
Superseded by #296

PR #297, opened 2023-02-08
Add bandit scan to pre-commit hooks

sandrobonazzola:
Adding [Bandit](https://bandit.readthedocs.io/en/latest/) scan to pre-commit hooks.

Signed-off-by: Sandro Bonazzola <sbonazzo@redhat.com>

### DIFF
--- a/.bandit.yml
+++ b/.bandit.yml
@@ -1,0 +1,4 @@
+---
+# https://bandit.readthedocs.io/en/latest/plugins/b101_assert_used.html?highlight=assert_used
+assert_used:
+  skips: ['tests/test_*.py', 'tests/plugins/test_*.py', 'tests/plugins/disable_test_trac.py']
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -39,6 +39,14 @@ repos:
             setup.py
           )$
 
+  - repo: https://github.com/PyCQA/bandit
+    rev: "1.7.4"
+    hooks:
+      - id: bandit
+        args:
+          - --configfile
+          - .bandit.yml
+
   - repo: https://github.com/pre-commit/pre-commit-hooks
     rev: "v4.4.0"
     hooks:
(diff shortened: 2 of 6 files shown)


bandit.yml
```diff
@@ -0,0 +1,3 @@
+---
+assert_used:
```
kwk, line 2: I suggest to rename the `bandit.yml` to `.bandit.yaml` for consistency reasons with the rest of the config files.

I'd also put a link to `assert_used` in the file so it is easier to understand where this file belongs to and where to find other configuration options that you might want to look into.

```suggestion
# https://bandit.readthedocs.io/en/latest/plugins/b101_assert_used.html?highlight=assert_used
assert_used:
```

kwk (changes requested):
Thank you for adding this security scanner. But we cannot enable it unless the pre-commit issues that it finds are dealt with. There are clearly false-positives when it comes to accidentally shared secrets. Here's [an example bandit finding](https://github.com/psss/did/pull/297/checks#step:4:134) for a piece of code that I've contributed recently:

```
>> Issue: [B107:hardcoded_password_default] Possible hardcoded password: 'token'
   Severity: Low   Confidence: Medium
   CWE: CWE-259 (https://cwe.mitre.org/data/definitions/259.html)
   Location: did/base.py:478:0
   More Info: https://bandit.readthedocs.io/en/0.0.0/plugins/b107_hardcoded_password_default.html
477	
478	def get_token(
479	        config: dict,
480	        token_key: str = "token",
481	        token_file_key: str = "token_file") -> str:
482	    """
```

I suggest to add exceptions or find another solution to similar findings.

kwk (commented):
Thank you addressing my concern but it gets quite noisy and I'm completely uncertain if this is really what we want: everytime the word token is used, we turn off security by given a comment `# nosec`. Not only is this noise but maybe bad if there was really a security problem. I mean what happens the next `token` function is implemented? Do we disable security there as well? That's quite unfortunate, don't you think?

PR #301, opened 2023-02-20
Address packit warning on `metadata` key

kwk:
To see the warning, go to an SRPM build log and search for this text:

```
WARNING The 'metadata' key in jobs is deprecated and can be removed. Nest config options from 'metadata' directly under the job object.
```

To reproduce the problem locally, use `packit srpm`.


psss (approved):
Thanks for the cleanup!

PR #303, opened 2023-03-31
Give a reasonable error for unreachable wiki url

psss:
Prevent traceback when wiki server is not reachable. Fix BZ#2118329.

### DIFF
--- a/did/plugins/wiki.py
+++ b/did/plugins/wiki.py
@@ -19,7 +19,7 @@ xmlrpc api endpoint::
 
 import xmlrpc.client
 
-from did.base import Config, ConfigError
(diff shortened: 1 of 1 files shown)


lukaszachy (commented):
How about catching also xmlrpc errors? 
This patch works for 'unreachable' however when I used example config ..

```
 ERROR  <Fault 1: 'This moin wiki does not allow xmlrpc method calls.'>

* Updates on wiki test: 0 changes of 0 pages
```

IMO it should handle that as well

psss:
Yeap, makes sense, fixed in ec6ba65e.

lukaszachy (approved):

PR #312, opened 2023-08-09
Support defining subitems for header and footer

### LINKED ISSUE #268: did fails to interpret `\n` in the config file
Perhaps I'm "misusing" `did` but before I run my daily reports I update the `Next` section of the footer. I would like to update it so that `did` can produce something like this:

- Morning plans
- Afternoon plans
- Evening plans

I inject the different strings into the config file, separated by `\n`, but `did` doesn't interpret these. How can I achieve this?

thanks

psss:
Convert escaped new lines into real new lines to allow printing custom subitems in header and footer.

Fix #268.

### DIFF
--- a/did/plugins/footer.py
+++ b/did/plugins/footer.py
@@ -9,6 +9,20 @@ Config example::
     type = footer
     next = Plans, thoughts, ideas...
     status = Status: Green | Yellow | Orange | Red
+
+New line characters can be used in the definition as well. In this way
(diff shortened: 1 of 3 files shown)


daobrien:
I've tested this quite a lot on my config and it works as expected.

sandrobonazzola (approved):
LGTM

lukaszachy (approved):

PR #313, opened 2023-08-16
Remove readfp and add dependency python3-feedparser

qiankehan:


psss (approved):
Thanks for both fixes! Looks good.

PR #314, opened 2023-08-24
add first WIP for allowing to get stats about created releases

### LINKED ISSUE #310: allow to report created github releases
Hello,

this is a great tool, thank you for creating it! :)

it would be great if it would be possible to track and report github releases that have been created by my account. 

Thank you for consideration! :)

szaimen:
Close https://github.com/psss/did/issues/310

### DIFF
--- a/did/plugins/github.py
+++ b/did/plugins/github.py
@@ -8,6 +8,7 @@ Config example::
     url = https://api.github.com/
     token = <authentication-token>
     login = <username>
+    project = <project1, project2>
 
 Optionally the search query can be limited to repositories owned
 by the given user or organization. You can also use the full name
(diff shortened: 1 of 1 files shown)


szaimen:
Hi @psss can you please give me same additional pointers where I need to modify the code so that it starts working? Thanks in advance! :)

psss:
Thanks for the draft! You just need to include the new stats in the `GitHubStats` class at the bottom of the file:

https://github.com/psss/did/blob/a3a49fa91076f3d3b7e5aa089bd13144a7b47009/did/plugins/github.py#L244-L266

Also make sure, that you correctly extract the list of projects from the `project` config value. You get inspiration here:

https://github.com/psss/did/blob/a3a49fa91076f3d3b7e5aa089bd13144a7b47009/did/plugins/github.py#L235-L239

Let me know if anything else is not clear.

did/plugins/github.py
```diff
@@ -9,6 +9,7 @@
     url = https://api.github.com/
     token = <authentication-token>
     login = <username>
+    project = <project1, project2>
```
psss, line 12:
```suggestion
    project = <project1>, <project2>
```

Let's make it a bit more clear that comma is a project separator here.

did/plugins/github.py
```diff
@@ -217,6 +218,16 @@ def fetch(self):
...
+class ReleasesCreated(Stats):
+    """ Releases created """
+
+    def fetch(self):
+        for project in self.user.projects:
+            log.info("Searching for releases created by {0} in {1}".format(self.user, project))
+            query = "repos/{0}/releases?q=author:{1}+created:{2}..{3}".format(
+                project, self.user.login, self.options.since, self.options.until)
+            self.stats = [
+                Releases(release) for release in self.parent.github.search(query)]
```
psss, line 230: Is seems that you need to define the `Release` class as well. It could be similar to the `Issue` class above.

psss (changes requested):
Nice kick-off! Thanks! In order to make it work we need the `Release` class though plus enabled the new stats under the `GitHubStats` class.

szaimen:
A new PR that implements this was opened here: https://github.com/psss/did/pull/472

PR #322, opened 2023-09-11
Replace deprecated calls to `warn`

sandrobonazzola:
Fixes:
```console
DeprecationWarning: The 'warn' method is deprecated, use 'warning' instead
```

when running with
```bash
export PYTHONWARNINGS="always::DeprecationWarning"
```


psss (approved):
Thanks for fixing this!

PR #326, opened 2023-09-22
Add markdown format output to pagure plugin

sandrobonazzola:
Add markdown format output to pagure plugin.


psss (approved):
Thanks for the improvement!

PR #328, opened 2023-09-25
Fix Zammad plugin by filtering tickets after search

come-nc:
The API query is not searching correctly and is returning more tickets that it should, so we loop on them and get articles from the API to filter which one were actually updated in the time range.

This reduces false positive a lot.

### DIFF
--- a/did/plugins/zammad.py
+++ b/did/plugins/zammad.py
@@ -19,6 +19,8 @@ import urllib.error
 import urllib.parse
 import urllib.request
 
+from datetime import datetime
+
 from did.base import Config, ReportError, get_token
 from did.stats import Stats, StatsGroup
 from did.utils import listed, log, pretty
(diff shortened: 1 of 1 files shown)


did/plugins/zammad.py
```diff
@@ -61,7 +63,25 @@ def search(self, query):
...
+        log.debug("Zammad query: {0}".format(url))
+        try:
+            request = urllib.request.Request(url, headers=self.headers)
+            response = urllib.request.urlopen(request)
+            log.debug("Response headers:\n{0}".format(
+                str(response.info()).strip()))
+        except urllib.error.URLError as error:
+            log.debug(error)
+            raise ReportError(
+                "Zammad search on {0} failed.".format(self.url))
```
psss, line 83: This part seems to be copy & pasted from the `search()` method. What about separating the common code to prevent duplication? There could be a common `search()` method used by both `get_articles()` and `get_tickets()`. What do you think, @come-nc?

psss (dismissed):
Thanks for improving this! Looks good, just two minor suggestions. Consider enabling `pre-commit` to catch code style errors early, see also the [develop](https://fmf.readthedocs.io/en/latest/contribute.html#develop) section.

sandrobonazzola:
@come-nc can you please rebase and address @psss comments?

come-nc:
> @come-nc can you please rebase and address @psss comments?

Rebased and fixed what I could, but it still complains the line is too long. I can rename the var `updated_at` into `u` but that’s kind of stupid?

sandrobonazzola (approved):
The plugin itself would benefit of some additional exception handling but can be done in separate patch.

PR #330, opened 2023-09-26
Update instructions for the `google` plugin

KwisatzHaderach:
Google has redesigned the APIs page and now the process is a little bit different.


psss (approved):
Thanks much for updating the instructions!

PR #331, opened 2023-09-29
Pass plugin configuration to koji `ClientSession`

sandrobonazzola:
pass plugin section configuration to `koji.ClientSession` as option, allowing to pass options like `no_ssl_verify = True` (useful for koji private instances running with self signed certificates).


psss (approved):
Looks good, thanks!

PR #337, opened 2023-11-14
Retry connecting to GitLab API on error

sandrobonazzola:
On connection error, retry to connect to the GitLab API. Found this usefull with https://gitlab.freedesktop.org/ which showed several connection resets for my case.


psss (approved):
Thanks for improving this! Looks good, added just a few couple of minor adjustments in f149820.

PR #340, opened 2023-11-16
Allow skipping events that are not reportable

xsuchy:
e.g. lunch break, focus block, deadline reminders

### DIFF
--- a/did/plugins/google.py
+++ b/did/plugins/google.py
@@ -11,6 +11,7 @@ Config example::
     client_secret = <client_secret>
     apps = calendar,tasks
     storage = ~/.did/google-api-credentials.json
+    skip = ["Lunch break", "Status deadline"]
 
 Make sure you have additional dependencies of the google plugin
 installed on your system::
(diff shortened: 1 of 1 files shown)


did/stats.py
```diff
@@ -22,6 +22,7 @@ class Stats(object):
     option = None
     dest = None
     parent = None
+    skip = None
```
psss, line 25: This should not be needed. The `skip` option belongs to configuration and it doesn't have to be defined here.

did/plugins/google.py
```diff
@@ -301,6 +304,7 @@ def __init__(self, option, name=None, parent=None, user=None):
             apps = [app.lower() for app in split(config["apps"])]
         except KeyError:
             apps = DEFAULT_APPS
+        self.skip = config.get("skip")
```
psss, line 307:
```suggestion
        self.skip = config.get("skip", [])
```

To make it working when no `skip` is present.

psss (changes requested):
Thanks for the improvement! Just two suggestions.

xsuchy:
Updated.

psss (approved):
Thanks for the adjustments, now looks good. Just modified the [commit message](https://did.readthedocs.io/en/latest/contribute.html#commits) for a consistent changelog.

PR #352, opened 2024-01-04
Implement `transition` stats for the `jira` plugin

lukaszachy:
Should I make it more generic? As in 'JiraTransitioned' ? In theory one could have config with one project counting 'Release Pending' and another where 'Integration' is counted as the user's role is different

Config like
```
[jira-where-developer]
project = DEVEL
transition_state = 'Integration'

[jira-where-qe]
project = QE
transition_state = 'Release Pending'
```

Seems that states and workflows can be too different

### DIFF
--- a/did/plugins/jira.py
+++ b/did/plugins/jira.py
@@ -29,7 +29,7 @@ token_name
     Name of the token to check for expiration in ``token_expiration``
     days. This has to match the name as seen in your Jira profile.
 
-transition_to
+transition_status
     Name of the issue status we want to report transitions to.
     Defaults to ``Release Pending`` (marking "verified" issues).
 
@@ -413,10 +413,10 @@ class JiraTransition(Stats):
         log.info(
             "[%s] Searching for issues transitioned to '%s' by '%s'",
             self.option,
-            self.parent.transition_to,
+            self.parent.transition_status,
             self.user.login or self.user.email)
         query = (
-            f"status changed to '{self.parent.transition_to}' "
+            f"status changed to '{self.parent.transition_status}' "
             f"and status changed by '{self.user.login or self.user.email}' "
             f"after {self.options.since} before {self.options.until}"
             )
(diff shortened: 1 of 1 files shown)


psss (approved):
Thanks for the improvement! The custom `verified_status` field looks good to me. Works nicely, I only ran into a problem when `login` contained `@`. Should be fixed by d339504.

psss:
For reference, the reported error was:

    Failed to fetch jira issues for query 'status changed to 'Release Pending' and status changed by psplicha@redhat.com after 2024-01-08 before 2024-01-15'.
    The reason was 'Bad Request' and the error was 'Error in the JQL Query: The character '@' is a reserved JQL character. You must enclose it in a string or use the escape '\u0040' instead. (line 1, character 67)'.

jharuda:
@lukaszachy Your implementation makes sense, generally speaking. How about specifying important states by `list` variable in the config? I am just thinking how to make it more general for different Jira projects.

The idea of this issue https://github.com/psss/did/issues/364 is to have for example something like this in the config:
```
[jira]
...
states=['Release pedning', 'In progress']
...
```
Default (without `states` variable in the config) behavior can be the same as it is now, which means generating only `updates` section to hide different states in the did output.

lukaszachy:
Or like `project in (PROJECT_NAME) AND "QA Contact" in (USER_NAME) AND status changed to ("Release Pending") during ("2024/05/13", "2024/05/17")`.

I guess I should update PR.

sandrobonazzola:
Can you please rebase?

sandrobonazzola:
pre-commit is failing on:

```
************* Module did.plugins.jira
did/plugins/jira.py:419:12: C0209: Formatting a regular string which could be an f-string (consider-using-f-string)
did/plugins/jira.py:426:28: C0209: Formatting a regular string which could be an f-string (consider-using-f-string)
did/plugins/jira.py:574:21: C0209: Formatting a regular string which could be an f-string (consider-using-f-string)
```

psss:
> Or like `project in (PROJECT_NAME) AND "QA Contact" in (USER_NAME) AND status changed to ("Release Pending") during ("2024/05/13", "2024/05/17")`.

This sounds very nicely flexible!

> I guess I should update PR.

I'd rather suggest to handle this in a separate pull request. We could have both target status (for basic use cases) and transition query (for more complex ones).

In order to make the naming a bit more clear I've renamed the key to `transition_status` in d05b116, for the advanced query we can use `transition_query` if that makes sense.

PR #355, opened 2024-02-09
Drop usage of distutils

sandrobonazzola:
Replace `distutils.util.strtobool` with a local copy of the function following the
[Migration advice](https://peps.python.org/pep-0632/#migration-advice) due to `distutils` module deprecation.

### DIFF
--- a/did/utils.py
+++ b/did/utils.py
@@ -185,6 +185,30 @@ def shorted(text, width=MAX_WIDTH):
     return "\n".join(lines)
 
 
+def strtobool(val):
+    """
+    Convert a string representation of truth to true (1) or false (0).
+    Reimplemented following instructions at
+    https://peps.python.org/pep-0632/#migration-advice
+    This is a copy of
+    https://github.com/pypa/distutils/blob/ee021a1c58b43607ccc75447159bd90f502c6bea/distutils/util.py#L340
+    Which is under MIT license.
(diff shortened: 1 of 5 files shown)


did/utils.py
```diff
@@ -187,6 +187,27 @@ def shorted(text, width=MAX_WIDTH):
     return "\n".join(lines)
 
 
+def strtobool(val):
+    """
+    Convert a string representation of truth to true (1) or false (0).
+    Reimplemented following instructions at
+    https://peps.python.org/pep-0632/#migration-advice
```
lukaszachy, line 194: Could you please add note that the code is an exact copy from  https://github.com/pypa/distutils/blob/ee021a1c58b43607ccc75447159bd90f502c6bea/distutils/util.py#L340

That code is under MIT license while all other `did` code is GPL so I think we should make it explicit (not a lawyer though).

lukaszachy (changes requested):

did/utils.py
sandrobonazzola: Done, also added a copy of MIT license.

lukaszachy (approved):

psss:
I don't think it's really worth it introducing an extra license just because of such a basic functionality. Let's do our custom simple implementation in #394.

PR #357, opened 2024-02-14
Avoid multiple queries to gitlab with no events

sandrobonazzola:
GitLab plugin was accessing GitLab API multiple times if the events set was empty. Now it accesses the API only once.


psss (approved):
Looks good, thanks for the improvement!

PR #365, opened 2024-05-09
Add token authentication to confluence plugin

sandrobonazzola:
It's now possible to use token authentication for connecting to Confluence.
Page list is now also available in markdown format.


sandrobonazzola:
Inclued via #390

PR #367, opened 2024-05-31
Added markdown support to google plugin

sandrobonazzola:
Added markdown support to google plugin


sandrobonazzola:
Included via #390

PR #368, opened 2024-07-08
Fix KeyError with user query result

waynesun09:
The return object could be dictionary when failed auth, which could cause the KeyError exception. Fix it by also catch the KeyError.


psss (approved):
Looks good, thanks!

PR #369, opened 2024-07-15
public_inbox: Fix email threads without an available root message 

mripard:
Hi,

@stefano-garzarella and @javierm have been reporting on the public-inbox plugin https://github.com/psss/did/pull/329 that they were having crashes.

It turns out that lore doesn't always manage to archive all the messages and we might find emails with a valid in-reply-to header, but can't retrieve the message from the ML archive.

This was partly due to an improper type annotation, so this PR fixes both the type annotations and the actual bug they were both facing.

### DIFF
--- a/did/plugins/public_inbox.py
+++ b/did/plugins/public_inbox.py
@@ -145,7 +145,7 @@ class PublicInbox(object):
         log.warn("Couldn't find message root")
         return None
 
-    def __get_thread_root(self, msg: Message) -> typing.Optional[Message]:
+    def __get_thread_root(self, msg: Message) -> Message:
         log.debug("Looking for thread root of message %s" % msg.id())
         if msg.is_thread_root():
             log.debug("Message is thread root already. Returning.")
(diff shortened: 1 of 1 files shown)


stefano-garzarella:
Thanks! Tested with my setup and this PR fixes the issue!

javierm:
Hi, @mripard :smile:.

I'm afraid I haven't reported anything; don't even know this project :thinking:. Wrong username, maybe?

mripard:
Yeah, sorry, I meant @martinezjavier

sandrobonazzola:
Can you please rebase?

did/plugins/public_inbox.py
```diff
@@ -143,8 +144,8 @@ def __get_msgs_from_mbox(self, mbox: mailbox.mbox) -> list[Message]:
 
         return msgs
 
-    def __fetch_thread_root(self, initial_msg: Message) -> Message:
-        msg_id = initial_msg.id()
+    def __fetch_thread_root(self, msg: Message) -> typing.Optional[Message]:
```
sandrobonazzola, line 147: using `msg` instead of `initial_msg` triggers error:

```
did/plugins/public_inbox.py:157:12: R1704: Redefining argument with the local name 'msg' (redefined-argument-from-local)
```

as msg is overridden by:

```python
 for msg in self.__get_msgs_from_mbox(mbox):
            if msg.is_thread_root():
            ...
```

is the overriding something you wanted?

mripard: You're right. I dropped the commit

PR #372, opened 2024-09-27
Allow to skip SSL verification in bugzilla plugin

sandrobonazzola:
Recently Red Hat bugzilla SSL certificate is failing verification. In order to allow querying bugzilla while this issue is getting fixed, allowing to skip SSL certificate validation.

Depends on #355


sandrobonazzola:
included via #390

PR #377, opened 2024-10-18
Avoid fetching thousands of pages from GitLab API

neverpanic:
When you comment on an issue in a project that has thousands of issues, or operate on a merge request in a project with thousands of MRs, did attempts to fetch them all to identify the ID of the one you interacted with.

This is OK for projects that are small in size, but becomes a huge issue for large projects such as https://gitlab.com/gitlab-org/gitlab, which has 218k issues.

Check for the number of pages that would be fetched and avoid loading them all if there are more than 20 pages. This will lead to issue and merge request IDs missing from the report, but it will allow the report to finish within your lifetime.


sandrobonazzola:
Can you please rebase?

neverpanic:
@sandrobonazzola Done, sorry for the delay.

sandrobonazzola (approved):
I'm approving as this solve a problem hitting people contributing to big projects. Just a note for follow up patches:

- Existing tests don’t cover:
  - The “over 20 pages” skip path.
  - get_project_issue / get_project_mr returning None and the resulting "unknown" in output.
- I've the feeling that having `MergeRequest.__init__`   still calling `super().__init__(data, parent, set_id=None)` will end up with `get_project_issue()` called again at some point, which can trigger loading all project issues (again with the same 20-page check).  There should be a way to avoid this.

pcahyna:
Is there a way to turn this off? And/or to configure the 20 pages limit? I believe I have projects where the limit would be exceeded, yet it is preferable to wait longer than to missing data in the report.

PR #379, opened 2025-01-01
Fetch all created issues correctly in `pagure`

mikelolasagasti:
Original query reports 6 created issues that are not closed:

https://pagure.io/api/0/user/mikelo2/issues?assignee=false&created=2024-01-01..2025-01-01

While the new one reports 169 created issues:

https://pagure.io/api/0/user/mikelo2/issues?status=all&created=2024-01-01..2025-01-01


psss (approved):
Thanks for the fix, @mikelolasagasti!

PR #392, opened 2025-04-15
Use custom `user-agent` in the `public_inbox` plugin

mripard:
The lore.kernel.org mailing-list archive recently started using the Anubis proxy to filter out AI scrap-bots.

Unfortunately, it seems to be somewhat based on the User-Agent, and the requests default one will get rejected. This means that the public_inbox will now be completely broken.

This can be bypassed by setting a custom User-Agent though, which restores the expected behaviour, so let's do so.

### DIFF
--- a/did/plugins/public_inbox.py
+++ b/did/plugins/public_inbox.py
@@ -16,12 +16,15 @@ import mailbox
 import tempfile
 import typing
 import urllib.parse
+from importlib.metadata import version
 
 import requests
 
 from did.base import Config, Date, ReportError, User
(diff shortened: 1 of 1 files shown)


eballetbo:
LGTM, I can confirm that with that patch the public-inbox plugin works again.
Thanks Maxime for this fix.

psss (approved):
Thanks much for the fix, Maxime! Confirmed that the plugin now works fine. However, for larger queries it fails, but that seems to be a different problem, filed as #393.

PR #403, opened 2025-05-30
Add proxy support for github and gitlab plugins

qiankehan:


sandrobonazzola:
Isn't it enough to export `https_proxy` in the console? Any reason for enabling proxy only for these 2 plugins instead of using proxy for all of them?

qiankehan:
@sandrobonazzola This patch will provide the proxy by site while https_proxy is the overall proxy for all sites in did.  
github and gitlab are often blocked by [GFW](https://en.wikipedia.org/wiki/Great_Firewall), so I add proxy for them.  
I could add proxy for other plugins if needed.

PR #404, opened 2025-06-03
Fix until returning type

sandrobonazzola:
GitHub.until needs to return did.base.Date instead of datetime.date


sandrobonazzola:
@Felixoid  can you please review?

psss:
Ah, sorry, was so focused on finishing the last items before the release that I haven't noticed this one. Completely agree with the fix. I've covered this in #405 together with extending the test coverage a bit.

PR #418, opened 2025-07-09
Drop phabricator plugin

sandrobonazzola:
Effective June 1, 2021: Phabricator is no longer actively maintained (see <https://secure.phabricator.com/>).

Dropping support for it.


sandrobonazzola:
I've no strong opinion but doesn't look it's widely used: https://we.phorge.it/w/usage/companies/
@wkw do you mind sending a PR moving from phabricator to phorge if you're still using it?
I'm closing this PR given the thumbs up from @psss to your comment on not dropping the plugin.

PR #427, opened 2025-08-07
Modernize Google plugin authentication

### LINKED ISSUE #415: `oauth2client` deprecation
`oauth2client` is used in google plugin and it has been deprecated by Google.

More information on the deprecation here: <https://google-auth.readthedocs.io/en/master/oauth2client-deprecation.html>

Need to refactor the plugin to use `google-auth` and `google-auth-oauthlib` instead.

sandrobonazzola:
Google Plugin Updates:
- Replace deprecated oauth2client with google-auth libraries
- Update from httplib2 to modern google.auth.transport.requests
- Add proper credential file management with JSON format
- Improve OAuth2 flow with better error handling
- Add support for credential refresh and re-authorization
- Update documentation for current Google API setup requirements

Test Coverage Improvements:
- Test credential file loading/saving with various formats
- Test OAuth flow error scenarios and credential refresh
- Test configuration validation and runtime error cases
- Test event filtering and task logging functionality
- Increase test coverage
- All tests properly mock external dependencies

Generated by: Claude 3.5 Sonnet (Anthropic AI Assistant) Code modernization and comprehensive test suite generated based on:
- Analysis of coverage gaps
- Best practices for Google API authentication
- Existing codebase patterns and testing strategies

Fixes #415


tests/plugins/test_google.py
```diff
@@ -140,55 +121,490 @@ def execute(self):
...
+
+    with tempfile.NamedTemporaryFile(mode='w', suffix='.json', delete=False) as f:
+        json.dump(cred_data, f)
+        f.flush()
+        try:
+            result = google.load_credentials_from_file(f.name)
+            assert result is not None
+            assert result.expiry is None  # Should be None due to ValueError
+        finally:
+            os.unlink(f.name)
```
kwk, line 269: Where's the difference in behaviour here? The `token_expiry` seems to have no effect on the `try` part. Also, is it important that `2025-12-31T23:59:59` lies in the future? If so I highly recommend to generate this date. Otherwise we're having a not so happy new-year :)

tests/plugins/test_google.py
```diff
@@ -140,55 +121,490 @@ def execute(self):
...
+
+    Generated by Claude 3.5 Sonnet to improve test coverage.
+    """
+    config = """
+[general]
+email = test@example.com
+
+[google]
+type = google
+client_secret = test_secret
```
kwk, line 389: I remember that I've implemented [secrets in files a while ago](https://github.com/psss/did/commit/f9421783557352bf38a4517f92a97594c96b8164#diff-4ed1b9d8e0d251153412685ff2b39aeb8b3a04384da2b5d86aefd985a5761051R46-R48). I would highly appreciate if you could add a test that stores the client secret in a file.

tests/plugins/test_google.py
```diff
@@ -140,55 +121,490 @@ def execute(self):
...
+
+    stats = google.GoogleStatsBase("test")
+    stats.parent = parent
+    stats._events = None  # pylint: disable=protected-access
+
+    events = stats.events
+    if events:
+        assert len(events) == 2
+        assert events[0].summary == "Meeting"
+        assert events[1].summary == "Important call"
```
kwk, line 507: IMHO the check for `if events` is wrong. The test should always run the assert. Otherwise you might silently ignore if you have skipped the wrong events. I think this Claude person is generating only correct python. But in a test I think the check is simply wrong.

tests/plugins/test_google.py
```diff
@@ -140,55 +121,490 @@ def execute(self):
...
+        ]
+
+    stats = google.GoogleStatsBase("test")
+    stats.parent = parent
+    stats._tasks = None  # pylint: disable=protected-access
+
+    with patch('did.plugins.google.log') as mock_log:
+        tasks = stats.tasks
+        if tasks:
+            assert len(tasks) == 2
```
kwk, line 528: Same here, why first check `tasks`?

kwk (changes requested):
Thank you for this PR. I have no way to test this. And I think the generated tests could be optimized. Sometimes the documentation of those tests is very poor and the code can accidentally hide real errors as I've described in my comments.

PR #448, opened 2026-02-23
Support Jira Cloud instance

lukaszachy:
Assisted-by: Claude Code

### DIFF
--- a/did/plugins/jira.py
+++ b/did/plugins/jira.py
@@ -267,8 +267,7 @@ class Issue():
                     "expand": expand})
             # Use new /search/jql endpoint for Jira Cloud
             # (required as of May 2025)
-            # See: https://developer.atlassian.com/changelog/
-            # #CHANGE-2046
+            # https://developer.atlassian.com/changelog/#CHANGE-2046
             search_endpoint = (
                 "search/jql" if stats.parent.is_jira_cloud else "search")
             current_url = (
@@ -435,7 +434,7 @@ class JiraStats(Stats):
 
             if not users:
                 raise ReportError(
-                    f"No user found for email {self.user.email} in Jira Cloud"
+                    f"No user found for email '{self.user.email}' in Jira Cloud."
                     )
 
             # Return the accountId of the first matching user
@@ -992,7 +991,7 @@ class JiraStatsGroup(StatsGroup):
             if self._session is not None:
                 return self._session
 
-            # Do not set it to self._sesson until it is fully ready
+            # Do not set it to self._session until it is fully ready
             _session = requests.Session()
             # Disable SSL warning when ssl_verify is False
             if not self.ssl_verify:


daobrien:
@sandrobonazzola @psss 

Apologies if this is not the best place to ask. I didn't see a Discussions section.

We're using Jira cloud now and I can't get the new config to work. The two main issues appear to be:
- Failed to connect to Jira at https://redhat.atlassian.net
- Skipping <Future at 0x7fa62b9f2b50 state=finished raised ReportError> due to Jira authentication failed.

~~~
[jira]
type = jira
url = https://redhat.atlassian.net/
auth_type = basic
auth_username = daobrien@redhat.com
auth_password = ***
api_version = 3
~~~

The section used to be called [jira-issues] and that worked. I don't really think that's the problem at all.
The `auth_password` is the API token that I created in the Jira cloud instance.

I have another section in the config file for Trello, and that works fine. It seems to be just changing to Jira cloud that's causing the problems. Did I miss something or get something wrong? Thanks for any help.

psss:
@daobrien, could you please try my [example config above](https://github.com/psss/did/pull/448#issuecomment-4098157188)? You should use `token` instead of `auth_password`. This should be also [documented](https://did.readthedocs.io/en/latest/plugins.html#module-did.plugins.jira). We just need to tidy up the sections a bit to make it more clear to the users.

daobrien:
> @daobrien, could you please try my [example config above](https://github.com/psss/did/pull/448#issuecomment-4098157188)? You should use `token` instead of `auth_password`. This should be also [documented](https://did.readthedocs.io/en/latest/plugins.html#module-did.plugins.jira). We just need to tidy up the sections a bit to make it more clear to the users.

@psss afaict my config is the same as yours apart from using `token`, which I've just updated. Unfortunately now I get a different error:
```ERROR  Skipping section jira due to error: `auth_password` or `auth_password_file` must be set in the [jira] section.```

Currently looking into the docs...

psss:
Do you have the latest `did` installed? The fresh [fedora update](https://bodhi.fedoraproject.org/updates/?search=did) on the way but you can also use the [did copr repo](https://copr.fedorainfracloud.org/coprs/psss/did/).

daobrien:
> Do you have the latest `did` installed? The fresh [fedora update](https://bodhi.fedoraproject.org/updates/?search=did) on the way but you can also use the [did copr repo](https://copr.fedorainfracloud.org/coprs/psss/did/).

Yes just installed it a couple of hours ago on F42. (Feel free to use Slack if that's easier/faster.)
Source RPM  : did-0.22-1.fc42.src.rpm

So, not the absolute latest. I'll add the COPR repo and try that.

daobrien:
> > Do you have the latest `did` installed? The fresh [fedora update](https://bodhi.fedoraproject.org/updates/?search=did) on the way but you can also use the [did copr repo](https://copr.fedorainfracloud.org/coprs/psss/did/).
> 
> Yes just installed it a couple of hours ago on F42. (Feel free to use Slack if that's easier/faster.) Source RPM : did-0.22-1.fc42.src.rpm
> 
> So, not the absolute latest. I'll add the COPR repo and try that.

Ok, that did it. Thanks 👍

psss:
> Ok, that did it. Thanks 👍

Good! :+1:

PR #449, opened 2026-02-24
Add type hints to zammad plugin

sandrobonazzola:
Add type hints to zammad plugin

### DIFF
--- a/did/plugins/zammad.py
+++ b/did/plugins/zammad.py
@@ -17,9 +17,10 @@ import json
 import urllib.error
 import urllib.parse
 import urllib.request
-from datetime import datetime
+from datetime import date, datetime
+from typing import Any, Optional, cast
 
-from did.base import Config, ReportError, get_token
+from did.base import Config, ReportError, User, get_token
 from did.stats import Stats, StatsGroup
 from did.utils import listed, log, pretty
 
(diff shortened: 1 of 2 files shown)


did/plugins/zammad.py
```diff
@@ -35,7 +36,7 @@ class Zammad():
     """ Zammad Investigator """
     # pylint: disable=too-few-public-methods
 
-    def __init__(self, url, token):
+    def __init__(self, url: str, token: Optional[str]) -> None:
```
kwk: I think we did have this before, right? But I cannot tell what reason we had not to use the more explicit and shorter syntax. The `str | None` syntax was introduced in Python 3.10 which is something we can expect I think. 

```suggestion
    def __init__(self, url: str, token: str | None) -> None:
```

> `Optional[X]` is equivalent to `X | None` (or `Union[X, None]`).
> [...]
> Changed in version 3.10: `Optional` can now be written as `X | None`. 
Source: https://docs.python.org/3/library/typing.html#typing.Optional

kwk:
> I agree the new notation is shorter but it requires 3.10 and will fail to work on CentOS Stream 9 and derivative as 3.9 is the default there.

Ack, then please only comment on the other suggestions I have made, not regarding the shorter syntax. I've resolved the ones that fall into the 3.9 category.

PR #452, opened 2026-03-17
Implement initial support for `forgejo` stats

sandrobonazzola:
Report created issues and pull requests.

Other cases requires deeper investigation on the API.


psss (approved):
Works nicely! Thanks much for implementing this.

PR #460, opened 2026-03-27
Support `--width 0` for unlimited output

### LINKED ISSUE #459: RFE: option to disable output truncation
When `did` outputs long text (e.g. long PR titles, Jira summaries, etc), it truncates lines with `...` based on the `--width` setting (default: `79`). There is currently no way to disable truncation entirely.                                                                                                                                

We could use `--width 0` to mean "no truncation".

ktdreyer:
Don't truncate lines with `...`

Co-Authored-By: Claude Opus 4.6 <noreply@anthropic.com>

Fixes: #459

### DIFF
--- a/did/utils.py
+++ b/did/utils.py
@@ -202,7 +202,7 @@ def shorted(text: str, width: int = MAX_WIDTH) -> str:
     lines = []
 
     for line in text.split("\n"):
-        if len(line) <= width:
+        if width == 0 or len(line) <= width:
             lines.append(line)
         else:
             # Remove any word after first overlapping non-word character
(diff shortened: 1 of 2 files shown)


ktdreyer:
Apologies, apparently this already works in `main`!

did/utils.py
```diff
@@ -202,7 +202,7 @@ def shorted(text: str, width: int = MAX_WIDTH) -> str:
     lines = []
 
     for line in text.split("\n"):
-        if len(line) <= width:
+        if width <= 0 or len(line) <= width:
```
sandrobonazzola: Why <= 0?  Aligning with the PR description, it should be `== 0` .

sandrobonazzola (changes requested):

did/utils.py
ktdreyer: Agreed. I've fixed this.

sandrobonazzola (approved):
LGTM

PR #470, opened 2026-05-19
fix(zammad): handle flat list response from search API

skjnldsv:
## Summary

Newer Zammad instances return a flat list of ticket objects from the
search endpoint instead of the `{"assets": {"Ticket": {...}}}` enveloppe.

## Fix

Detect the response type:
- list → build `{str(id): ticket}` dict directly
- dict → existing `assets.Ticket` unwrap path

Both paths produce the same structure consumed by `TicketsUpdated.fetch()`.

## Traceback

```
  ...
  File ".../did/plugins/zammad.py", line 114, in fetch
    for _, ticket in self.parent.zammad.search(query).items():
                     ~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^
  File ".../did/plugins/zammad.py", line 63, in search
    result = self.perform_search(query)["assets"]
             ~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^
TypeError: list indices must be integers or slices, not str
```

Tested against a Zammad instance runing the newer API format, confirmed working.

### DIFF
--- a/did/plugins/zammad.py
+++ b/did/plugins/zammad.py
@@ -60,11 +60,16 @@ class Zammad():
                 f"Zammad search on {self.url} failed.") from error
 
     def search(self, query: str) -> dict:
-        result = self.perform_search(query)["assets"]
-        try:
-            result = result["Ticket"]
-        except KeyError:
-            result = {}
+        result = self.perform_search(query)
+        if isinstance(result, list):
+            # Newer Zammad API returns a flat list of ticket objects
+            result = {str(ticket["id"]): ticket for ticket in result}
+        else:
+            result = result["assets"]
+            try:
+                result = result["Ticket"]
+            except KeyError:
+                result = {}
         log.debug("Result: %s fetched", listed(len(result), "item"))
         log.data(pretty(result))
         return result


skjnldsv:
cc @come-nc

come-nc (commented):
I guess that works.
We actually do not need the keys so it would be cleaner to turn the map into list for older instances instead (and adapt the for to ignore keys in TicketsUpdated::fetch).

The diff I use locally is:
```diff
diff --git a/did/plugins/zammad.py b/did/plugins/zammad.py
index 78d331c..15f942e 100644
--- a/did/plugins/zammad.py
+++ b/did/plugins/zammad.py
@@ -59,12 +59,8 @@ class Zammad():
             raise ReportError(
                 f"Zammad search on {self.url} failed.") from error
 
-    def search(self, query: str) -> dict:
-        result = self.perform_search(query)["assets"]
-        try:
-            result = result["Ticket"]
-        except KeyError:
-            result = {}
+    def search(self, query: str):
+        result = self.perform_search(query)
         log.debug("Result: %s fetched", listed(len(result), "item"))
         log.data(pretty(result))
         return result
@@ -111,7 +107,7 @@ class TicketsUpdated(Stats):
         self.stats = []
         since = self.options.since.date
         until = self.options.until.date
-        for _, ticket in self.parent.zammad.search(query).items():
+        for ticket in self.parent.zammad.search(query):
             for article in self.parent.zammad.get_articles(ticket["id"]):
                 updated_at = datetime.fromisoformat(
                     article["updated_at"].replace('Z', '+00:00')).date()
```

I did not create a PR from it because I did not know how to support both old and new.

DerDreschner (approved):

PR #471, opened 2026-06-22
gitlab: return all events

lukaszachy:
get_all_results was omit by mistake

Assisted-by: Claude Code


lukaszachy:
Failing test is `/tests/github/issues`, which should cover github not gitlab -> unrelated to this change?

lukaszachy:
Changed it to look as all other `_get_gitlab_api_list` calls.

PR #477, opened 2026-07-14
Add a new release notes section to the docs

psss:
Create a new docs/releases.rst page covering the last three releases and the upcoming 0.24. Add the page to the docs table of contents.

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -19,6 +19,7 @@ Table of Contents
 
     Overview <overview>
     Install <install>
+    Releases <releases>
     Config <config>
     Examples <examples>
     Plugins <plugins>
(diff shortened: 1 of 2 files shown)


kwk (changes requested):
@psss you've been creating a lot of AI commits lately.

I suggest that the AI puts links to the relevant PRs or commits next to each update in a release so it is easy to verify. I followed the PRs that went into did and I think this mostly covers it but it would still be nice if your AI could add the links. Otherwise the review for us humans takes longer than the time to make the PR. Thanks.

psss:
> I suggest that the AI puts links to the relevant PRs or commits next to each update in a release so it is easy to verify.

Good idea! Added in ce1bcd5.

psss:
@kwk, sounds ok/better now?

kwk (approved):
@psss yes, much better. Thank you!

PR #479, opened 2026-07-31
Add the `uv` installation method

### LINKED ISSUE #436: update readme for uv installation method
The readme mentions `pip install did`. A more modern solution would be `uv tool install did`.

lntutor:
Fixes #436.

Adds a dedicated `uv` installation section using `uv tool install did`, while retaining the existing package, virtual environment, and plugin-extra instructions.

Validation:
- `pre-commit run codespell --files docs/install.rst`
- `git diff --check`

The full pre-commit environment could not finish locally because its existing mypy hook dependency set failed while building `gssapi` on macOS.


lntutor:
Current head `f698c02cd3e7aef52f6d4d9e6caf4af0fad98612` is mergeable, and the visible current-head checks are in a clean state: Read the Docs is green, the Packit RPM builds are green, the Testing Farm jobs are green, and the `osh-diff-scan` result is neutral rather than a failing gate. There are no review threads or review-body comments open on the PR. If this documentation addition still matches the intended direction, it looks ready for maintainer review/merge when convenient.

docs/install.rst
```diff
@@ -26,6 +26,14 @@ This will bring dependencies for all core plugins as well.
 __ https://copr.fedoraproject.org/coprs/psss/did/
 
 
+UV
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Install did as an isolated command-line tool using uv::
+
+    uv tool install did
```
psss, line 34: Would it make sense to also shortly mention `uv tool run did` as the newly installed script is not automatically added to `PATH`? Might be useful for new `uv` users. @lntutor, what do you think?

